PR #25, opened 2019-10-21
Improve Chart.js benchmark by skipping label rotation

benmccann:
I'm working on the Chart.js documentation to add some performance tips and have [suggested they add this tip](https://github.com/chartjs/Chart.js/pull/6585). I don't think Chart.js will ever be as fast as uPlot, but it should be at least as fast as Highcharts, etc.

Setting `maxRotation: 0` allows Chart.js to skip calculating how much to rotate the labels to fit them on the chart and just makes them all horizontal.

`autoSkipPadding` mostly just makes it look a little nicer by putting some spacing between labels

### DIFF
--- a/bench/Chart.js.html
+++ b/bench/Chart.js.html
@@ -103,7 +103,9 @@
 								distribution: 'series',
 								ticks: {
 									source: 'auto',
-									autoSkip: true
+									maxRotation: 0,
+									autoSkip: true,
+									autoSkipPadding: 75
 								}
 							}],
 							yAxes: [{


leeoniya:
hey @benmccann, thanks for the PR.

i can merge this but the net effect (if there is one) appears to be noise-level. not all that surprising since so few labels need to be measured here.

are you seeing anything different on your end?

benmccann:
Ah, yeah, it looks like you're right. I was working off a locally build version of Chart.js to debug perf issues and it looks like that version is taking longer for some reason

If you go back to Chart.js 2.6.0 this benchmark runs a lot faster. I've been trying to track down the regressions. I think we fixed [the main one](https://github.com/chartjs/Chart.js/commit/6bc6630b87e6e56ece8bbd9721d7c42d3a0e6e86), so hopefully Chart.js 2.9.0 won't be quite as terrible

leeoniya:
do you want to hold off and PR this with 2.9 then?

benmccann:
Whatever's easier for you. I don't think this option hurts and would probably specify it either way. At some point in the next couple weeks I'll follow up to bump the version number

PR #31, opened 2019-10-26
Upgrade to Chart.js 2.9.0

benmccann:
The results before and after are really variable, but this looks like it's around 25-33% faster.

The two largest issues that Chart.js still has in terms of speed compared to uPlot are accepting varied input (parsing, linear and timeseries support in time scale, etc.) and animation support (which is still costly even when off due to the way the code is structured).

### DIFF
--- a/bench/Chart.js.html
+++ b/bench/Chart.js.html
@@ -11,7 +11,7 @@
 		<h2 id="wait">Loading lib....</h2>
 
 		<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment.min.js"></script>
-		<script src="https://www.chartjs.org/dist/2.8.0/Chart.min.js"></script>
+		<script src="https://www.chartjs.org/dist/2.9.0/Chart.min.js"></script>
 		<script src="https://www.chartjs.org/samples/latest/utils.js"></script>
 
 		<canvas id="chart1"></canvas>
@@ -105,7 +105,8 @@
 									source: 'auto',
 									maxRotation: 0,
 									autoSkip: true,
-									autoSkipPadding: 75
+									autoSkipPadding: 75,
+									sampleSize: 1
 								}
 							}],
 							yAxes: [{


benmccann:
Chart.js works with several different date libraries. It uses moment by default, but I'm not sure it's the fastest. The results differ so much on my machine when I run just a handful of times manually. I'm not sure if you have an automated way to run 100x or anything like that, but I think it might be faster to use Luxon instead. E.g. replace the scripts at the top of the page with:

```
		<script src="https://cdn.jsdelivr.net/npm/luxon@1.15.0"></script>
		<script src="https://cdn.jsdelivr.net/npm/chart.js@2.9.0"></script>
		<script src="https://cdn.jsdelivr.net/npm/chartjs-adapter-luxon@0.1.1"></script>
		<script src="https://www.chartjs.org/samples/latest/utils.js"></script>
```

leeoniya:
thanks! i'll test this soon.

> accepting varied input (parsing, linear and timeseries support in time scale, etc.)

this doesnt make too much sense to me. i should be able to provide 100k datapoints in your native format and avoid 100% of any parsing overhead. hopefully you don't runtime-test each datapoint to handle non-uniform arrays! checking items[0] datatype up-front and assuming the rest are the sams should be basically free.

leeoniya:
on my laptop (and now Chrome 78.0.3904.70) i'm seeing a good improvement:

1230ms -> 860ms

the variability is pretty high to tell how much Luxon helps, but looks like it shaves an additional ~15ms.

i'll post the results once i test on the other machine (and verify nothing's changed since Chrome 77)

leeoniya:
oops,  i merged this earlier than intended. i can't update the bench table until i get to the same machine tomorrow or tues.

benmccann:
No problem. 2.9.1 was just released, so might be worth upgrading to that version when you run the benchmark since it's the latest. (I expect it would be slightly faster though potentially negligible difference)

leeoniya:
some initial numbers on the test machine (now also on Chrome 78):

uPlot - 43ms
Highcharts - 640ms
Chart.js 2.9.1 - 980ms

there aren't the absolute lowest numbers for each lib. for instance, Chart.js hits 1050ms as frequently as low 980s rarely 960ms. the variability in runs of all libs seems to be about ±4% from the numbers i try to post.

i'll re-test the others later today and update the table.

PR #39, opened 2019-10-31
fix bug where s is undefined

seanyusa:
This fixes an issue where on line 128 (now 129) s is undefined.

I don't know exactly how you are building, so I'll leave that up to you!


leeoniya:
thanks!

PR #41, opened 2019-11-02
Tags each scale with the min and max series to avoid scaleReset when …

### LINKED ISSUE #40: optimize scale invalidation when toggling a series
right now when a series is toggled, the scale associated with it is invalidated (along with all paths using that scale) and recomputed and paths redrawn.

by additionally keeping track of which series bumped the final scale min/max, we can avoid invalidating the scale when toggling a series that never affected the its associated scale. this would avoid recomputing the min/max of the scale and regenerating the other paths, which would provide a healthy improvement by avoiding most of the work. this is especially apparent when you have a lot of data-heavy series on the same scale.

what's involved is adding a prop or two to the scale itself for tagging which series indices pegged the min/max values by keeping track inside of `setScales()`, then adding a check in `toggle()` to avoid unnecessary invalidation.

jeancharles-roger:
Here is a proposition to close #40 with updated derived-scale for test.

I'm not sure with the equality test.


jeancharles-roger:
I'm still learning pull requests ... I accidentally closed this one 😅

leeoniya:
no big deal. open as many as you need 🤣

PR #44, opened 2019-11-02
Upgrade Chart.js

benmccann:
I found another pretty good performance boost there: https://github.com/chartjs/Chart.js/pull/6671


leeoniya:
healthy improvement. though maybe somewhat worse on mem usage. i suspect this will all get sorted out by v3, by the looks of it.

i'll probably add another column to the bench soon for retained heap, not just peak. lots of spread there, too.

PR #46, opened 2019-11-03
Caches min and max for each series for optimized toggle

### LINKED ISSUE #40: optimize scale invalidation when toggling a series

jeancharles-roger:
Adds min and max for each serie on the current view to avoid resetting scale if not necessary when toggled.

Fixes #40.


dist/uPlot.cjs.js
```diff
@@ -521,6 +521,8 @@ function uPlot(opts, data) {
 		if (!(key in scales)) {
 			scales[key] = {
 				auto: true,
+				rawMin:  inf,
+				rawMax: -inf,
```
leeoniya: i'd prefer to re-compute these on demand from the series' min/max caches. it's cheap and reduces cache maintenance in more places than necessary.

src/uPlot.js
```diff
@@ -132,6 +132,8 @@ export default function uPlot(opts, data) {
 		if (!(key in scales)) {
 			scales[key] = {
 				auto: true,
+				rawMin:  inf,
+				rawMax: -inf,
```
leeoniya: i'd prefer to re-compute these on demand from the series' min/max caches. it's cheap and reduces cache maintenance in more places than necessary.

src/uPlot.js
```diff
@@ -432,7 +438,13 @@ export default function uPlot(opts, data) {
 		return [round6(incrRoundUp(scaleMin, incr)), scaleMax];
 	}
 
-	// TODO: add ability to recompute & invalidate only specific scales. e.g. for series toggle
+	function getSerieMinMax(serie, data, _i0, _i1) {
+		if (serie.minMax == null) {
+			serie.minMax = getMinMax(data, _i0, _i1)
```
leeoniya: to keep it uniform with scales structures let's keep the properties as separate. `s.min` and `s.max`

src/uPlot.js
```diff
@@ -793,7 +814,10 @@ export default function uPlot(opts, data) {
 				toggleDOM(ip, onOff);
 			}
 
-			resetScale(s.scale);
+			// reset scale if current serie is actually the min or max bound
+			let sc = scales[s.scale]
+			if (!s.minMax || sc.rawMin >= s.minMax[0] || sc.rawMax <= s.minMax[1])
```
leeoniya: i don't think `>=` and `<=` are correct. they should probably be strict `>` and `<`, otherwise this will have the same issue as the previous version, that it will reset the scale even if there is another series with the same min or the same max. right?

leeoniya: as i mentioned in https://github.com/leeoniya/uPlot/issues/40#issuecomment-549065923, i think we need to make sure that the toggled series is exceeds all other mins or all other maxes of all series shown on the same scale. if a min or max is the same as any another one, it shouldn't invalidate.

src/uPlot.js
jeancharles-roger: Nice, I'll do that

src/uPlot.js
jeancharles-roger: Actually, the test is right. It's counter intuitive because the `sc.rawMin` is computed with `s.min` so `sc.rawMin > s.min` won't be true (unless the series was hidden and the toggle shows it).
But considering that I won't be storing `rawMin` anymore I'll just recompute it with (or without) the new series bounds. It should be simpler to understand and maintain.

src/uPlot.js
jeancharles-roger: It's done

src/uPlot.js
leeoniya: try it with this dataset:

```js
[
  [1,2,3,4],
  [1,2,1,2],
  [2,1,2,1],
]
```

toggling one should never invalidate the scale or paths.

src/uPlot.js
```diff
@@ -806,6 +803,16 @@ export default function uPlot(opts, data) {
 		(isArr(idxs) ? idxs : [idxs]).forEach(i => {
 			let s = series[i];
 
+			// computes current scale min and max 
+			var scaleMin = inf
+			var scaleMax = -inf
```
leeoniya: `let`

src/uPlot.js
```diff
@@ -439,10 +440,12 @@ export default function uPlot(opts, data) {
 	}
 
 	function getSerieMinMax(serie, data, _i0, _i1) {
```
leeoniya: in English, "Series" is the correct term for both single and multiple series.

src/uPlot.js
```diff
@@ -815,8 +822,9 @@ export default function uPlot(opts, data) {
 			}
 
 			// reset scale if current serie is actually the min or max bound
-			let sc = scales[s.scale]
-			if (!s.minMax || sc.rawMin >= s.minMax[0] || sc.rawMax <= s.minMax[1])
+			// if the toggled serie was the lower bound then s.min == scaleMin (same for max)
+			// the test then must be `s.min <= scaleMin` (or `s.max >= scaleMax`).
+			if (s.min <= scaleMin || s.max >= scaleMax)
```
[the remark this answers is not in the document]
leeoniya: when i toggle a single series using the data below, it resets the scale and rebuilds the path of the remaining series. this should not happen.

```js
[
  [1,2,3,4],
  [1,2,1,2],
  [2,1,2,1],
]
```

src/uPlot.js
jeancharles-roger: Thanks for the example. I rethink it 😅.

src/uPlot.js
jeancharles-roger: `let` doesn't declare a constant, sorry I did know

src/uPlot.js
leeoniya: i'm looking through more of this task and realizing this may not have been `good first issue` D:

i'm writing a patch myself to see how many more places need to be touched. so you maybe hold off on any additional work for now. sorry 😭

jeancharles-roger: Ok no problem. I tried 🤣, but it was a good way to get in the code and learn somethings (like let doesn't declare a constant 🤓). I'll probably try again

Thanks a lot for your reviews.

dist/uPlot.cjs.js
```diff
@@ -1217,10 +1207,21 @@ function uPlot(opts, data) {
...
+					scaleMax = max(scaleMax, cs.max);
+				}
+			});
+			// snap scale
+			var sc = scales[s.scale];
+			var minMax = sc.range(scaleMin, scaleMax);
+			
+			// compare previous scale to new one
+			if (sc.min !== minMax[0] || sc.max !== minMax[1]) 
 				{ resetScale(s.scale); }
```
jeancharles-roger: Here I recompute the scale bounds from the cache min/max in series including the snap. If the scaled changed it needs to be reset. There is probably a better way to do it but it doesn't work :)

leeoniya: i'm a bit embarrassed that i didnt notice this myself  😬. but yes, that's an issue!

leeoniya: maybe still keeping the raw min/max cache at the scale level will be good, to reduce a bunch of extra loops. we'll see where i end up.

leeoniya:
thanks for your work @jeancharles-roger, the final patch is here: https://github.com/leeoniya/uPlot/commit/6a14df9014fded07065767ef5a1e1972012f41db.

quick summary:

- scales did end up needing to be compared *after* snapping
- adding the min/max series caches allowed scale resetting/re-computing to become cheap
- scale invalidation is now not done directly but via manual invalidation of series' min/max caches and direct `series.path` nulling

this ended up making the whole flow much more sane/logical. scale recomutation will auto-recalc min/maxes if they're invalidated and canvas drawing will re-calc the paths if they're invalidated.

i hope this does not discourage you from contributing in the future. thanks again!

jeancharles-roger:
Great !. No worries I'll try again (maybe on something less ambitious 😉).
Thank you for your reviews I learn a lot.

Issue #43, opened 2019-11-02
Internationalization


benmccann:
You can use the `Intl` API to format times so that they render in other language-agnostic format. E.g. I work on a site that supports both English and Spanish and "October" would automatically show up as "Octubre" in my charts on that site because they're using Luxon which uses the browser's `Intl` api under the covers. You don't need to provide any translations as the browser has them built in. You could either use a library like Luxon or use the `Intl` API directly

leeoniya:
![3f2vzv](https://user-images.githubusercontent.com/43234/68074583-a3eba180-fd6a-11e9-98f4-660f551cd10e.jpg)

leeoniya:
i've been hoping to side-step these concerns by exposing full control over the axis and legend value gen. while uPlot ships with and exposes a fast English formatter [1], and exposes `axis.values` and `series.value` that provide the full story for all date formatting concerns. it's easy to override all of these and use Luxeon or whatever to fully customize the displayed values. i'm afraid of bloating the lib and creating undue maintenance burden by providing more than the core functions that are necessary for customization by the user.

i've always known that whatever label format i chose to ship as the default would only please 20% of uPlot's audience. even if i were to use the intl API directly, i don't see how it would be possible to avoid the snowball effect of offering and supporting a lot more.

[1] https://github.com/leeoniya/uPlot/blob/master/src/fmtDate.js

benmccann:
Yeah, I agree users will want ability to change formats. `Intl` api might provide a nice default, but even still I agree that there will be many requests to override and change the formatting.

leeoniya:
this is likely out of scope for a micro lib. as long as enough nuts and bolts are there to add support externally.

thanks for bringing it up tho!

PR #47, opened 2019-11-04
Added an option (divisors) that facilitates changing the division intervals

danyalejandro:
For our use case, we prefer to control the division intervals (governed by the axis.incrs variable). I believe other potential users might also want to tweak the intervals to their own data visualization needs.

This commit allows the user to specify a new optional property "divisors" for the x axis, where he can set new divisors for: "minute", "hour", "day", "month", "year" separately.

All properties are optional and will fall back to the default divisors when not provided.

Example configuration:

```js
var h = 3600;

axes: {
    x: {
        divisors: {
            day: [ h, 2*h, 3*h, 4*h, 6*h, 8*h ]
        }
    }
}
```


src/uPlot.js
```diff
@@ -236,12 +236,45 @@ export default function uPlot(opts, data) {
...
+	 * 
+	 * @param {object} axis element in the axes array
+	 * @param {boolean} isTime True if axis.type is "t"
+	 * @returns {array} Array of integers, as expected by axis.incrs
+	 */
+	function parseDivisors(axis, isTime) {
+		// If the user provided the option, parse it
+		if (axis.hasOwnProperty("divisors")) {
+			if (isTime) {
+				let inc = dec;
```
leeoniya: ?

leeoniya:
hey @danyalejandro 

custom `incrs` can already be passed via opts. what is the added benefit of this?

danyalejandro:
Yes, it could be done by the user as long as he reads the library's code, understands the meaning of the values and builds a new array of divisors that work. I tried to do it and it's too cumbersome to expect a normal user to do it, most data scientists won't be able to.

The added benefit is the improved usability in your library's API design.

leeoniya:
i would like to keep the bare minimum API necessary to get the job done.

for example, it was impossible previously to draw high/low bands; not difficult but *impossible*. so i added this because you made a valid case for it (thanks!). #43 Internationalization is another example of this - it was rejected because there is sufficient API surface to get it done without touching the core.

if you feel that your users cannot define the `incrs` in the format uPlot expects, then simply provide them a helper function which accepts `divisors` in a format you feel they would be happy with and return the incrs that uPlot expects. the result would be nearly identical to them:

```js
var h = 3600;

axes: {
    x: {
        incrs: parseDivisors({
            day: [ h, 2*h, 3*h, 4*h, 6*h, 8*h ]
        })
    }
}
```

leeoniya:
i'm gonna close this one, but i think a helper function that aids in customizing axis `incrs`, `ticks`, and `values` would indeed be a good thing to have as something like a `recipes` wiki page here. but it would have to be pretty powerful to really be useful, like ability to set proper value formatting breakpoints, etc. the reason i left all of this out of the core is because there's going to be 1M opinions about how it should behave.

let me know if there's something which is not available in the core to be able to do externally.

thanks!

leeoniya:
@danyalejandro 

you may be interested in a recent addition that adds a simplified API for configuring axis tick formatting. essentially you can now pass a config struct into `axis.values` instead of re-implementing the full function logic manually:

```js
values: [
  [3600 * 24 * 365,    "{YYYY}",               7    "{YYYY}",                       0],
  [3600 * 24 * 28,     "{MMM}",                7,   "{MMM}\n{YYYY}",                0],
  [3600 * 24,          "{M}/{D}",              7,   "{M}/{D}\n{YYYY}",              0],
  [3600,               "{h}{aa}",              4,   "{h}{aa}\n{M}/{D}",             0],
  [60,                 "{h}:{mm}{aa}",         4,   "{h}:{mm}{aa}\n{M}/{D}",        0],
  [1,                  "{h}:{mm}:{ss}{aa}",    4,   "{h}:{mm}:{ss}{aa}\n{M}/{D}",   0],
];
```

more notes on the format in the source: https://github.com/leeoniya/uPlot/blob/c9e9d51b9993cc1e606aec4fdaca512255bcede3/src/opts.js#L105-L122

in combination with `axis.space` and `axis.incrs`, this gives a lot of configuration power. the other piece is `axis.ticks`, but this one is quite complex and better left alone since it's what handles timezone and DST logic.

danyalejandro:
Definitely interested, thank you!

PR #50, opened 2019-11-07
Chart.js - use linear distribution of time

kurkle:
Use default `'linear'` distribution of time. Its faster.
`'series'` mode plots each point with equal distance between, `'linear'` uses the given time to determine distance.


leeoniya:
looks good, thanks!

PR #65, opened 2019-12-07
Add Chart.js 3 benchmark

benmccann:
It'd be cool to get this on the README if you're willing so we can see how we stack up

### DIFF
--- a/bench/Chart.js3.html
+++ b/bench/Chart.js3.html
@@ -106,7 +106,10 @@
 						parsing: false,
 						tooltips: {
 							mode: 'index',
-							intersect: false,
+							intersect: false
+						},
+						hover: {
+							axis: 'x'
 						},
 						scales: {
 							x: {
--- a/bench/Chart.js3.html
+++ b/bench/Chart.js3.html
@@ -1,0 +1,153 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8">
+		<title>Chart.js Demo</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1">
+
+		<link rel="stylesheet" href="style.css">
+		<style>
+			.chart-container {
+				position: relative;
+				margin: auto;
+				height: 600px;
+				width: 1920px;
+			}
+		</style>
+	</head>
+	<body>
+		<h2 id="wait">Loading lib....</h2>
+
+		<script src="https://cdn.jsdelivr.net/npm/luxon@1.15.0"></script>
+		<script src="https://www.chartjs.org/dist/master/Chart.min.js"></script>
+		<script src="https://cdn.jsdelivr.net/npm/chartjs-adapter-luxon@0.2.0"></script>
+		<script src="https://www.chartjs.org/samples/latest/utils.js"></script>
+
+		<div class="chart-container">
+			<canvas id="chart1"></canvas>
+		</div>
+
+		<script>
+			function round2(val) {
+				return Math.round(val * 100) / 100;
+			}
+
+			function round3(val) {
+				return Math.round(val * 1000) / 1000;
+			}
+
+			function prepData(packed) {
+				console.time('prep');
+
+				// epoch,idl,recv,send,read,writ,used,free
+
+				const numFields = packed[0];
+				packed = packed.slice(numFields + 1);
+
+				var cpu = Array(packed.length/numFields);
+				var ram = Array(packed.length/numFields);
+				var tcp = Array(packed.length/numFields);
+
+				for (let i = 0, j = 0; i < packed.length; i += numFields, j++) {
+					let date = packed[i] * 60 * 1000;
+					cpu[j] = {x: date, y: round3(100 - packed[i+1])};
+					ram[j] = {x: date, y: round2(100 * packed[i+5] / (packed[i+5] + packed[i+6]))};
+					tcp[j] = {x: date, y: packed[i+3]};
+				}
+
+				console.timeEnd('prep');
+
+				return [cpu, ram, tcp];
+			}
+
+			function makeChart(data) {
+				console.time('chart');
+
+				var ctx = document.getElementById('chart1').getContext('2d');
+				Chart.defaults.global.animation = false;
+				var color = Chart.helpers.color;
+				var cfg = {
+					data: {
+						datasets: [
+							{
+								label: 'CPU',
+								borderColor: window.chartColors.red,
+								data: data[0],
+								type: 'line',
+								pointRadius: 0,
+								fill: false,
+								lineTension: 0,
+								borderWidth: 1
+							},
+							{
+								label: 'RAM',
+								borderColor: window.chartColors.blue,
+								data: data[1],
+								type: 'line',
+								pointRadius: 0,
+								fill: false,
+								lineTension: 0,
+								borderWidth: 1
+							},
+							{
+								label: 'TCP Out',
+								borderColor: window.chartColors.green,
+								data: data[2],
+								type: 'line',
+								pointRadius: 0,
+								fill: false,
+								lineTension: 0,
+								borderWidth: 1
+							}
+						]
+					},
+					options: {
+						maintainAspectRatio: false,
+						parsing: false,
+						tooltips: {
+							mode: 'index',
+							intersect: false,
+						},
+						scales: {
+							x: {
+								type: 'time',
+								distribution: 'series',
+								ticks: {
+									source: 'auto',
+									maxRotation: 0,
+									autoSkip: true,
+									autoSkipPadding: 75,
+									sampleSize: 1
+								}
+							},
+							y: {
+								scaleLabel: {
+									display: true,
+									labelString: '%'
+								}
+							}
+						},
+					}
+				};
+
+				var chart = new Chart(ctx, cfg);
+
+				wait.textContent = "Done!";
+				console.timeEnd('chart');
+			}
+
+			let wait = document.getElementById("wait");
+			wait.textContent = "Fetching data.json (2.07MB)....";
+			fetch("data.json").then(r => r.json()).then(packed => {
+				wait.textContent = "Rendering...";
+				let data = prepData(packed);
+				setTimeout(() => makeChart(data), 0);
+			});
+		</script>
+
+		<p>
+			Code based on<br>
+			<a href="https://www.chartjs.org/samples/latest/scales/time/financial.html">https://www.chartjs.org/samples/latest/scales/time/financial.html</a><br>
+		</p>
+	</body>
+</html>


leeoniya:
i couldn't get this commit working (console errors).

but i locally compiled your `parsing` branch as of https://github.com/chartjs/Chart.js/pull/6814 and got 50% boost across the board vs `Chart.js.html`:

Chart.js.html
```
740ms Scripting
195ms System
117mb peak
 90mb retained
```

Chart.js3.html
```
383ms Scripting
 95ms System
 64mb peak
 46mb retained
```

great strides!

benmccann:
@leeoniya my `parsing` branch was committed to Chart.js `master`. Does this PR work for you now? If not, what are the console errors you get?

bench/Chart.js3.html
```diff
@@ -0,0 +1,153 @@
...
+
+				var chart = new Chart(ctx, cfg);
+
+				wait.textContent = "Done!";
+				console.timeEnd('chart');
+			}
+
+			let wait = document.getElementById("wait");
+			wait.textContent = "Fetching data.json (2.07MB)....";
+			fetch("data.json").then(r => r.json()).then(packed => {
```
benmccann: this line is a little annoying because it doesn't work on local. what do you think about changing all the samples to use `https://leeoniya.github.io/uPlot/bench/data.json` instead?

leeoniya: > it doesn't work on local

i usually test all the demos and benches with `http-server`

> what do you think about changing all the samples to use https://leeoniya.github.io/uPlot/bench/data.json instead?

i'm frequently on a tethered connection and pumping 2mb of json with caching disabled in devtools would be a no-go for me (both slow and i'd burn through my cap 10x faster).

benmccann:
@leeoniya was there anything else you wanted to do before merging this? Also, let me know if you have suggestions on the file name. It seemed a little funny to me, but I couldn't think of anything better

leeoniya:
> @leeoniya was there anything else you wanted to do before merging this?

no, this looks fine. i'll merge it when i have a bit more time for re-benching. i'm working on some uPlot API stuff that's higher in the queue. i see you may have more perf-related stuff coming up anyways: https://github.com/chartjs/Chart.js/pull/6822

leeoniya:
@benmccann i should add that i would be hesitant to merge an implementation that comes with a bunch of caveats and disabled features, specifically made to move up in the bench tables. the fact that the fastline controller excludes datapoint rendering is not a good sign - what if i zoom in on a region that's sparse enough to show datapoints, am i SOL if i chose this controller originally?

i realize i'm judging the featureset from a biased point of view, but i don't want implementors carving out modes in their libs just so they can say "look, we can do a `lineTo()` loop, too!".

curating benchmarks is hard.

benmccann:
Sure, that makes sense. Just to clarify, all the work I'm doing on improving performance is motivated primarily by my own use case. And this PR doesn't use the `fastline` implementation since it sets `type: 'line'` and not `type: 'fastline'`

I'm not entirely sure yet whether `fastline` is something I'll end up pursuing or merging or what form it would take if I did. It was kind of a proof of concept I wrote on a lark. I found a couple interesting things from doing it such as a [bug in the Chart.js code](https://github.com/chartjs/Chart.js/pull/6826) and also a change that would need to be made to the base controller to enable it. I think it's worth at least making some of these other supporting changes even if we don't pursue `fastline` because it would enable users to create their own controllers based off of `fastline`

The main performance benefit of `fastline` comes from disabling animations. The reason I didn't include datapoint rendering or spanGaps was just because I didn't need it in my own project. I don't think it'd be hard to support if someone wanted it. I just didn't have a reason to do that work for my own project

leeoniya:
the current non-minified version [1] works fine, but the minified version [2] gives me this:

image

[1] https://www.chartjs.org/dist/master/Chart.js
[2] https://www.chartjs.org/dist/master/Chart.min.js

benmccann:
Hmm. I don't get that. I did "Ctrl + F5" to try to clear the cache and it still worked for me. Can you try that to see if maybe it was just a problem with an old version?

leeoniya:
> Can you try that to see if maybe it was just a problem with an old version?

what kind of dev do you take me for? 🤣 

here's a fresh copy/pasta that fails for me: https://jsfiddle.net/80wabtvx/

benmccann:
> what kind of dev do you take me for?

haha. I'm stumped though as to what else it could be

This is really weird because the jsfiddle works for me

Screenshot from 2019-12-14 14-47-00

@kurkle @etimberg could you guys give the fiddle link above a try? I'm not sure why it'd be failing for @leeoniya and working for me

leeoniya:
fails on my phone too. maybe it's a cdn issue. can u attach the version u're getting? or content hash.

benmccann:
passes on my phone :-)  here's the file my browser's loading (I had to append `.txt` to the filename to get GitHub to allow me to upload it)

[Chart.min.js.txt](https://github.com/leeoniya/uPlot/files/3964488/Chart.min.js.txt)

leeoniya:
yes, it looks like a CDN issue. we're not getting the same versions.

i'm at home now (was in cafe before) and still have issues in both browsers. here's what i get served: [Chart.min.js.txt](https://github.com/leeoniya/uPlot/files/3964515/Chart.min.js.txt). mine is coming from `[2606:4700::6811:2035]:443` with the etag below. what's yours?

image.

benmccann:
Wow. I never would have guessed that. Here's mine

Screenshot from 2019-12-14 19-07-38

leeoniya:
so, any thoughts on how to move forward here? (it's still failing for me, btw).

benmccann:
Hmm. I wonder if you try on ipv4 if it will work?

@etimberg @kurkle do you guys know how https://chartjs.org/dist/master/Chart.min.js is hosted and deployed to Cloudflare? Is there anything you guys can think of to make sure the latest version is deployed and caching is disabled?

leeoniya:
> I wonder if you try on ipv4 if it will work?

it might, but that's irrelevet unless you can force all users of that link to disable ipv6.

benmccann:
Yeah, I'm still trying to figure out what would cause that, so just looking for any useful clues

leeoniya:
same issue on ipv4:

104.17.33.53:443
W/"5dd68267-2b1cf"

etimberg:
We didn't enable cloudflare, that must be something GitHub is doing for pages. When I test, I get

IP: 104.17.32.53:443
Etag: `W/"5df03279-2bcc6"`
cf-ray: 54636299aea4cab0-YYZ
x-served-by: `cache-bwi5050-BWI` (this is super interesting. I am in YYZ so not sure why the cache is going to BWI)

benmccann:
It's in the chartjs.org DNS settings, so I think we did configure it. I'll email @nnnick and see if there's anything he can do to help

benmccann:
@leeoniya can you try again? I think this might be fixed now. Response I got from Nick:

> I checked and the /dist files have a very high edge TTL, since a lot of people hotlink assets directly and usually the paths are canonical. Think tagged releases etc. Hadn’t considered the latest master branch, so I’ve updated to drop that to one hour for master branch.

leeoniya:
yep, that did it.

benmccann:
Woohoo!

leeoniya:
i'll re-bench this soon on Chrome 79.

meanwhile, Chrome just keeps getting slower and slower: https://github.com/krausest/js-framework-benchmark/issues/683

leeoniya:
current numbers:

```
uPlot
  42                console.timeEnd("chart")
  71,6,3,72         script/rend/paint/sys  (first render)
  20.1MB -> 3.6MB   heap (max -> retained)
  153,517,129,259   script/rend/paint/sys  (interact 10s)

Flot
  320
  192,5,5,306
  39.6MB -> 17.1MB

dygraphs
  170
  239,4,5,172
  113MB -> 64.1MB
  2365,260,316,328

Chart.js-next
  290
  368,6,4,81
  56.2MB -> 49.3MB
  3705,36,93,4553

CanvasJS
  321
  398,5,4,103
  47.9MB -> 37.4MB
  2030,714,350,445

LightningChart
  ---
  472,3,3,78
  67.2MB -> 21.4MB
  9725,25,47,98

jqChart
  455
  564,4,2,85
  141MB -> 97.3MB
  675,665,328,485

Highcharts
  ---
  687,7,2,72
  82.4MB -> 28.1MB
  1150,581,194,276

Chart.js
  680
  726,6,4,175
  113MB -> 90.8MB
  5564,5,13,4067

ECharts
  505
  778,5,9,1074
  182MB -> 124MB
  2472,51,52,7333

ApexCharts
  1265
  2398,26,152,60
  144MB -> 123MB
  1815,205,8119,57

ZingChart
  5858
  6007,9,1,80
  206MB -> 174MB

amCharts
  6915
  6953,37,12,102
  441MB -> 441MB
  4522,1237,2977,518
```

benmccann:
Cool! Is there anything else you wanted to do before merging this PR now that we have the new benchmark?

Thanks for all your help on this one! It was tricky to figure out that Cloudflare issue!

leeoniya:
> Is there anything else you wanted to do before merging this PR

i'm gonna pillage my markdown table builder from domvm [1] and maintain the benchmarks in a JSON file. i like the denser style of the non-html table (i need more columns) and formatting that manually is gonna get old quickly.

[1] https://github.com/domvm/domvm/blob/master/build.js#L234-L292

leeoniya:
still need to link the benches from the new table & re-check lib sizes.

benmccann:
cool. i'm curious what's the difference between `done` and `js` in the next col?

leeoniya:
`done` == `console.timeEnd("chart")`. this is just pure render time, excluding lib script parsing, JSON parsing & data prep. `js` the the total amount of js recorded in the perf timeline summary from page load to `done` (or somewhat after, assuming everything is settled).

benmccann:
Thanks for the detailed explanation! It's funny to me that "done" for Flot is greater than "js". I'd expect "js" to always be larger since it includes more. I probably would have chosen to order on the "js" column since data prep is normally done on the server side in my experience

leeoniya:
> I probably would have chosen to order on the "js" column since data prep is normally done on the server side in my experience

well, you can't exactly do data prep for libs that need native `Date` objects, so what then?

`js` *includes* data prep & script parsing. since i just record a full perf timeline, i cannot - without a lot of effort - include script parsing (which is relevent, because less js should be rewarded) but also exclude the data prep that immediately follows it.

in theory, `done` should be all that's needed (and for a lot of libs it is). but then there are cases where `done` fires before the chart is painted. in other cases there libs load data urls (highcharts actually caches a png image of the plot, which takes time but i think happens after `done`).

i was considering ordering everything by `js + sys`, but that doesn't match the real experience. i think Chart.js-next does better than Flot in that sorting order, but if you actually open the page, Flot is very obviously faster. it'll never be perfect; it is what it is.

also, i updated the lib sizes and you guys put on some weight: 239 -> 247, but i don't recall 8k worth of features landing 🤷‍♂ .

benmccann:
Yeah, if anything I thought we deleted code. We added babble, so that could be part of it. I'll have to finish the buble PR. I'm surprised it's that big though because it was showing as only 174k in that PR for me: https://github.com/chartjs/Chart.js/pull/6836

leeoniya:
i'm including all deps required to run the bench (besides what's inlined in the html) - so, `luxeon` in your case and `jQuery` for Flot.

Issue #27, opened 2019-10-24
range, point & line annotations


leeoniya:
http://dygraphs.com/gallery/#g/highlighted-weekends
http://dygraphs.com/gallery/#g/highlighted-region

also annotations along the axes/scales (vt or hz lines that mark some specific event in time or some level along a y axis).

series annotations [1] will probably be handled by #10:

[1] http://dygraphs.com/gallery/#g/annotations

mcintyre94:
Would this cover the functionality displayed in this Apex charts demo: https://apexcharts.com/javascript-chart-demos/line-charts/line-chart-annotations/?

leeoniya:
pretty much, though uPlot almost certainly won't be dealing with the the label positioning. i'll add the labels to the legend but leave anything fancier to be implemented externally.

leeoniya:
@mcintyre94

i think actually this should be possible to do outside the core via a plugin. uPlot provides everything you need for this, like translating scale values to coordinates and hooks to when the view or data changes.

take a look at the composable `tooltips` and `wheel-zoom` demos:

https://leeoniya.github.io/uPlot/demos/tooltips.html
https://leeoniya.github.io/uPlot/demos/wheel-zoom.html

 would you be interested in prototyping an annotations plugin?

leeoniya:
following up on #76...

a couple things to keep in mind:

i would prefer all text labels to be drawn in the dom since you get CSS styling, text wrapping and interactivity for free. the same thinking applies for the actual regions/lines/points - if they're on canvas, they're effectively static (without adding a bunch of extra code), but if they're drawn using dom elements, then we get a lot more for free, like hover interaction, on-demand labels & options for label placement libs that respect dom element bounds, etc.

the price we pay for drawing regions/lines/points in the dom is the inability to place them anywhere except above or below the full canvas (you cannot place them in between the grid and the series) - i'm not sure if this is a show-stopper for keeping everything in the dom - it might be since we'll want the text labels to be above everything, but i can see wanting the annotations themselves be between the grid and series.

another option would be to have the grid on a separate canvas, but this wastes a bunch of memory by allocating an entire canvas region just to draw some lines - which is not great for full-screen plots, e.g. 1920x1080.

leeoniya:
this is all now easy enough to do in userland that it can be delegated to a plugin.

highlighted-weekends remains a bit tricky because of the required timezone and dst accounting, but #119 exists for this specific case.

Issue #10, opened 2019-10-10
opt to show data points, labels or custom markers on lines


leeoniya:
eg ability to draw candlesticks at the datapoints.

https://news.ycombinator.com/item?id=21216316

PR #76, opened 2019-12-23
draw hooks

ldstein:
Hi @leeoniya,

This PR proposes adding `willPaint` and `didPaint` hooks to `Line`.

It will enable custom rendering onto the Canvas element, either before or after the Chart is drawn.

I added this functionality as an approach to add highlighted annotated regions of a Chart (BG color below, Text labels above). 

If you feel it's a worthwhile addition, I'll add some documentation to `readme.md` accordingly.

**Edit** I discovered I could achieve a similar result using overlayed/underlayed canvas / div elements and rely on `setScale`,  `setFocus` for notification of change. I'll leave the PR open regardless for your consideration.

image

### DIFF
--- a/demos/draw-hooks.html
+++ b/demos/draw-hooks.html
@@ -1,0 +1,257 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8">
+		<title>Draw Hooks</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1">
+
+		<link rel="stylesheet" href="../bench/style.css">
+		<link rel="stylesheet" href="../src/uPlot.css">
+		<style>
+			.uplot {
+				margin-top: 50px;
+			}
+		</style>
+	</head>
+	<body>
+		<script src="../dist/uPlot.iife.js"></script>
+		<script>
+
+            function addCrazyVisualsPlugin() {
+
+                // Source: https://stackoverflow.com/questions/21646738/convert-hex-to-rgba
+                function hexToRgbA(hex, a='1') {
+                    let c;
+                    if(/^#([A-Fa-f0-9]{3}){1,2}$/.test(hex)){
+                        c= hex.substring(1).split('');
+                        if(c.length== 3){
+                            c= [c[0], c[0], c[1], c[1], c[2], c[2]];
+                        }
+                        c= '0x'+c.join('');
+                        return 'rgba('+[(c>>16)&255, (c>>8)&255, c&255].join(',')+',' + a +')';
+                    }
+                    throw new Error('Bad Hex');
+                }
+
+                // Source: https://stackoverflow.com/questions/25837158/how-to-draw-a-star-by-using-canvas-html5
+                function drawStar(ctx, cx,cy,spikes=6,outerRadius=8,innerRadius=4){
+
+                    outerRadius = outerRadius * devicePixelRatio;
+                    innerRadius = innerRadius * devicePixelRatio
+
+                    var rot=Math.PI/2*3;
+                    var x=cx;
+                    var y=cy;
+                    var step=Math.PI/spikes;
+
+                    ctx.beginPath();
+                    ctx.moveTo(cx,cy-outerRadius)
+                    for(i=0;i<spikes;i++){
+                        x=cx+Math.cos(rot)*outerRadius;
+                        y=cy+Math.sin(rot)*outerRadius;
+                        ctx.lineTo(x,y)
+                        rot+=step
+
+                        x=cx+Math.cos(rot)*innerRadius;
+                        y=cy+Math.sin(rot)*innerRadius;
+                        ctx.lineTo(x,y)
+                        rot+=step
+                    }
+                    ctx.lineTo(cx,cy-outerRadius);
+                    ctx.closePath();
+                }
+
+                let medians = [];
+
+                // https://www.jstips.co/en/javascript/array-average-and-median/
+                function calculateMedians(u)
+                {
+                    medians = u.data.map((data) => {
+                        data = [...data];
+                        data.sort((a, b) => a - b);
+                        return (data[(data.length - 1) >> 1] + data[data.length >> 1]) / 2
+                    });
+                }
+
+                function drawBackgroundGradient(u)
+                {
+                    let {ctx} = u;
+                    let {width, height} = ctx.canvas;
+
+                    let gradient = u.ctx.createLinearGradient(0, 0, 0, height);
+                    gradient.addColorStop(0, "#666");
+                    gradient.addColorStop(1, "#000");
+
+                    ctx.save();
+                    ctx.fillStyle = gradient;
+                    ctx.fillRect(0,0,width,height);
+                    ctx.restore();
+                }
+
+                function drawSeriesMedian(u, i)
+                {
+                    // https://github.com/leeoniya/uPlot/issues/77
+                    i -= 1;
+
+                    let {ctx} = u;
+                    let {width, height} = ctx.canvas;
+                    let {label, color, min:minVal, max:maxVal, scale} = u.series.y[i];
+
+                    cy = u.valToPos(medians[i+1], scale) * devicePixelRatio;
+
+                    ctx.save();
+                    ctx.strokeStyle = hexToRgbA(color, 0.2);
+                    ctx.lineWidth = 50 * devicePixelRatio;
+                    ctx.filter = 'blur(6px)';
+
+                    ctx.beginPath();
+                    ctx.moveTo(0, cy);
+                    ctx.lineTo(ctx.canvas.width, cy);
+                    ctx.closePath();
+                    ctx.stroke();
+                    ctx.restore();
+                }
+
+                function logDrawGridAxis(u, a)
+                {
+                    // TODO - Think of something creative to do here
+                    let axis = u.axes[a];
+                    console.log("Draw Grid Axis", a, axis);
+                }
+
+                function enableBlurBeforeGridRender(u)
+                {
+                    let {ctx} = u;
+                    ctx.save();
+                    ctx.filter = 'blur(2px)';
+                }
+
+                function disableBlurAfterGridRender(u, axis)
+                {
+                    let {ctx} = u;
+                    // This is a bit buggy. There are some cases where the y axis will not be rendered
+                    // Resulting in the the canvas to be rendered with the previously applied blur
+                    if (axis == 'y')
+                        ctx.restore();
+                }
+
+                function drawPointsAsStars(u, i)
+                {
+                    // https://github.com/leeoniya/uPlot/issues/77
+                    i -= 1;
+
+                    let {ctx} = u;
+                    let {label, color, scale} = u.series.y[i];
+                    let yVals = u.data[i+1];
+
+                    ctx.save();
+                    ctx.fillStyle = color;
+                    yVals.forEach((val, i) => {
+                        // Can add additional check if visible on Canvas before drawing
+                        let cx = u.valToPos(u.data[0][i], 'x') * devicePixelRatio;
+                        let cy = u.valToPos(val, scale) * devicePixelRatio;
+                        drawStar(ctx, cx, cy);
+                        ctx.fill();
+                    });
+
+                    ctx.restore();
+                }
+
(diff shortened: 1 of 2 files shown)


leeoniya:
thanks for the PR, this is cool :)

there's actually a couple discussions to be had here.

1. generic paint/draw hooks
2. range annotations & this demo

**1. generic paint/draw hooks**

generally, i think having generic draw hooks is a good idea &  escape hatch. while i'm no stranger to will/did hook implementations [1], i think having this style just for draw hooks is not ideal. effectively, all of uPlot's current hooks are `did*` hooks. additionally, i think it would be better to have more than just two but to make it possible to draw into any series' "layer", not just above or below all of them. so maybe something like:

- `drawClear` fires after the canvas is cleared
- `drawGrid` fires after the grid is drawn
- `drawSeries: (self, i)` (fires after each series is drawn)
- `draw` fires after everything is drawn

rather than passing `ctx` into the hooks i'd rather follow the existing convention and simply expose `ctx` as `self.ctx` (since it's a singleton). hooks that accept extra params are the ones that need further context (like "which" in setSeries and setScale).

for this demo specifically, you should not need to replicate getXPos & getYPos - they're already exposed:

https://github.com/leeoniya/uPlot/blob/589d1dd2a17d7fdc33aef176b837678131b460c2/dist/uPlot.cjs.js#L1672-L1674

**2. range annotations & this demo**

i think that chart annotations provide sufficient data inference utility to be part of the core, so there's https://github.com/leeoniya/uPlot/issues/27 already open to discuss the possible API. i'll comment a bit further there.

[1] https://github.com/domvm/domvm#lifecycle-hooks

ldstein:
Thanks for the feedback, all valid points.

I updated to reflect your input:
- Removed `willPaint`, `didPaint`
- Added `drawClear`, `drawGrid`, `drawSeries`, `draw` hooks
- Updated demo html

`draw-hooks.html` demonstrates various effects which can be applied, such as adding blur to the grid:

image

Regarding **"you should not need to replicate getXPos & getYPos"**, there is a small offset issue when using `valToPos()` or `posToVal()` if the user's browser is set to a zoom level other than 100%.

Appears related to `valToPos()` using `cssWidth` instead of `canvas.width` when calculating position. Calculated correctly for DOM elements, but missing an offset when applied to canvas space. 

Hence why I copied `getXPos` & `getYPos` in the original demo :) Here is an example of the issue:

image

**Other Draw Event Candidates**
- `drawGridAxis` might be another valid hook worth considering.
- drawPath() could potentially fire `drawSeriesStroke`, `drawSeriesFill`

src/Line.js
```diff
@@ -717,7 +722,8 @@ export function Line(opts, data) {
 
 				ctx.translate(-offset, -offset);
 			}
-		});
+        });
+        fire("drawGrid");
```
leeoniya: let's move this up into `if (grid.show) {` above. it'd be weird if `drawGrid` fired when there's no grid to draw.

src/Line.js
```diff
@@ -536,8 +538,11 @@ export function Line(opts, data) {
 		});
 
 		series.forEach((s, i) => {
-			if (i > 0 && s.show)
-				drawPath(i);
+			if (i > 0 && s.show) {
+                drawPath(i);
+                fire("drawSeries", s, i);
```
leeoniya: this should just be `fire("drawSeries", i);` the `s` can already be pulled from `self.series.y[i+1]`

src/Line.js
```diff
@@ -742,10 +748,12 @@ export function Line(opts, data) {
...
 
-		ctx.clearRect(0, 0, can[WIDTH], can[HEIGHT]);
-		drawAxesGrid();
-		drawSeries();
-		didPaint = true;
+        ctx.clearRect(0, 0, can[WIDTH], can[HEIGHT]);
+        fire("drawClear");
+        drawAxesGrid();        
+        drawSeries();
+        fire("draw");
```
leeoniya: i'd probably move this down 1 line (below `didPaint = true;`)

src/Line.js
ldstein: Maybe `fire("drawAxis", axis)` would be more appropriate in that case? 

Can access grid from `axis.grid`

leeoniya:
> Regarding "you should not need to replicate getXPos & getYPos", there is a small offset issue when using valToPos() or posToVal() if the user's browser is set to a zoom level other than 100%.
>
> Appears related to valToPos() using cssWidth to calculate position. It calculates correctly for DOM space, but missing a scaling modifier when applied to canvas space.
>
> Hence why I copied getXPos & getYPos in the original demo :) Here is an example of the issue:

you need to multiply the position by `devicePixelRatio`. this solves the issue on high res displays as well as higher zoom levels:

```js
let yPos0 = u.valToPos(maxVal, 'y') * devicePixelRatio;
let yPos1 = u.valToPos(minVal, 'y') * devicePixelRatio;
```

> Other Draw Event Candidates

honestly, i don't feel the value of those is that great.

please switch to tabs spacing and clean up all the superfluous whitespace changes, then maybe squash the commits. otherwise this looks good!

demos/draw-hooks.html
```diff
@@ -0,0 +1,231 @@
...
+                        setStartRenderTime,
+                        drawBackgroundGradient,
+                        enableBlur
+                    ],
+                    drawGrid: [
+                        disableBlur
+                    ],
+                    drawSeries: [
+                        enableBlur,
+                        drawSeriesBoundary,
```
leeoniya: i didn't actually realize what this was doing until i disabled it. until that point it looked like it was just part of the a blurred gradient background.

src/Line.js
leeoniya: yep, good catch.

leeoniya: err, sorry. i meant `fire("drawGrid", i)`. the axes are not really "drawn" on the canvas (they're in the dom). so let's stick with canvas terminology.

ldstein: I changed this to `fire('drawGridAxis', a)` for each axis. If 1 or more axis are drawn, `drawGrid` will be fired. 

Allows hooking in directly after all axis have been drawn and before series is drawn.

demos/draw-hooks.html
```diff
@@ -0,0 +1,237 @@
...
+                        x=cx+Math.cos(rot)*innerRadius;
+                        y=cy+Math.sin(rot)*innerRadius;
+                        ctx.lineTo(x,y)
+                        rot+=step
+                    }
+                    ctx.lineTo(cx,cy-outerRadius);
+                    ctx.closePath();                    
+                }
+
+                function setCtxProperties(ctx, settings)
```
leeoniya: i'm not a canvas expert, but isn't this effectively https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/save?

demos/draw-hooks.html
ldstein: Yeah, not a particularly great example. It's more apparent once you zoom in (then you get a bounding box around the visible points).

src/Line.js
```diff
@@ -540,7 +540,7 @@ export function Line(opts, data) {
 		series.forEach((s, i) => {
 			if (i > 0 && s.show) {
                 drawPath(i);
-                fire("drawSeries", s, i);
+                fire("drawSeries", i-1);
```
leeoniya: right now i'm passing the raw internal `i` here [1], so this should not be adjusted to assume differently. whether this is ultimately a good decision is probably another topic, but let's keep it uniform in this PR.

[1] https://github.com/leeoniya/uPlot/blob/master/src/Line.js#L929

leeoniya: i've opened #77 to figure this out.

leeoniya:
thanks for contributing! i did some cleanup commits.

ldstein:
Thanks @Leeyon,

I meant to comment earlier, but you Christmas kind of took priority :)

Having experimented a bit now, the real-life requirements I have can be accomplished with just the `drawClear` and `draw` hooks. 

It may be enough for 80% of use cases and maintains the "black box" around Uplot's internal drawing routines:

--------------------------------------------
"drawClear" Any custom drawing here
--------------------------------------------
uPlot drawing methods (controlled via options) 
--------------------------------------------
"draw" Any custom drawing here
--------------------------------------------

As you already mentioned,  #27 potentially competes with 'drawSeries'.
Leaving out 'drawSeries' also means no need to solve #77 right away.

Also note there is a bug in the demo I provided:

When `drawClear` fires, I enable a canvas blur filter (in preparation of drawing the grid). I then turn this off by when `drawGrid` is fired and checking if the axis being drawn is `y`.

However, `drawGrid` for the y axis does not fire all the time (for example, when zooming in a great distance). The result is the blur filter is never be turned off, so the entire chart is blurred.

A `drawGrids/drawAllGrids` (which is always fired regardless of how many grids were drawn) would allow hooking into directly before the series are drawn and would solve this particularly issue. 

Again, I don't think I'd miss it if the `drawGrid` hook didn't exist. I could always configure the grid with semi-transparent colors and use `drawClear` to do something fancy.

Food for thought.

ldstein:
Thanks @Leeyon,

I meant to comment earlier, but you Christmas kind of took priority :)

Having experimented a bit now, the real-life requirements I have can be accomplished with just the `drawClear` and `draw` hooks. 

It may be enough for 80% of use cases and maintains the "black box" around Uplot's internal drawing routines:

--------------------------------------------
"drawClear" Any custom drawing here
--------------------------------------------
uPlot drawing methods (controlled via options) 
--------------------------------------------
"draw" Any custom drawing here
--------------------------------------------

As you already mentioned,  #27 potentially competes with 'drawSeries'.
Leaving out 'drawSeries' also means no need to solve #77 right away.

Also note there is a bug in the demo I provided:

When `drawClear` fires, I enable a canvas blur filter (in preparation of drawing the grid). I then turn this off by when `drawGrid` is fired and checking if the axis being drawn is `y`.

However, `drawGrid` for the y axis does not fire all the time (for example, when zooming in a great distance). The result is the blur filter is never be turned off, so the entire chart is blurred.

A `drawGrids/drawAllGrids` (which is always fired regardless of how many grids were drawn) would allow hooking into directly before the series are drawn and would solve this particularly issue. 

Again, I don't think I'd miss it if the `drawGrid` hook didn't exist. I could always configure the grid with semi-transparent colors and use `drawClear` to do something fancy.

Food for thought.

leeoniya:
> I meant to comment earlier, but Christmas kind of took priority :)

of course, merry christmas, btw!

> As you already mentioned, #27 potentially competes with `drawSeries`.

i think this `draw-hooks` demo (not `drawSeries`) "competes" with #27, (and also #10). but while those features are absent, it's a good thing to show; once they're added, the this demo will need to be changed to something that doesn't overlap those APIs.

> Having experimented a bit now, the real-life requirements I have can be accomplished with just the `drawClear` and `draw` hooks.

but it's pretty cheap to have the others, too. and without them it's impossible to draw into each series' layer and allow those to be visually occluded by later series.

> Omitting `drawSeries` also means no need to solve #77 right away.

#77 also affects `setSeries` hooks, so it's a priority for me before 1.0. the current plan is to get rid of the x/y layers and just have flat arrays for `series` and `axes`. for series, it makes perfect sense because they always need 1:1 mapping to the data array. for axes, `side` is really the thing that determines where they're drawn, so `x` and `y` nesting them is just unnecessary indirection.

when both become arrays, the rules are simple. `axes[0]` defaults to 'x' and `axes[1..N]` default to `y` options. same goes for `series`. this means the opts will always be:

```js
{
  series: [
    {},
    {color: "red"},
  ],
  axes: [
    {},
    {values: ...},
  ]
}
```

sadly, it's not a beautiful API, but it resolves the `i` situation without any mental gymnastics or additional conventions. that's kind of where i'm at with that.

> Also note there is a bug in the demo I provided:

feel free to submit any improvements, but the demos don't need to 100% bug free, more just starting points for more experimentation. i'd rather have imperfect but simple demos rather than demos which handle all edge cases but are more difficult to grok. i think leaving a comment about the bug is fine, especially if the alternative is another 20 lines of code or more state-tracking complexity.

ldstein:
No worries, sounds good. Thanks again for all your effort, and Merry Christmas to you too!

Fiv5:
Nice conversation, I learn a lot!

PR #78, opened 2019-12-29
Minor optimization

benmccann:
* Move declaration of `s` outside `map` function, so it only needs to be done once
* Clarify that ticks contain only label values

### DIFF
--- a/src/opts.js
+++ b/src/opts.js
@@ -129,6 +129,7 @@ export const _timeAxisStamps = timeAxisStamps([
 export function timeAxisVals(tzDate, stamps) {
 	return (self, ticks, space) => {
 		let incr = ticks[1] - ticks[0];
+		let s = stamps.find(e => incr >= e[0]);
 
 		// these track boundaries when a full label is needed again
 		let prevYear = null;
@@ -143,7 +144,6 @@ export function timeAxisVals(tzDate, stamps) {
 			let diffYear = newYear != prevYear;
 			let diffDate = newDate != prevDate;
 
-			let s = stamps.find(e => incr >= e[0]);
 			let stamp = s[2] == 7 && diffYear || s[2] == 4 && diffDate ? s[3] : s[1];
 
 			prevYear = newYear;


leeoniya:
> Move declaration of `s` outside `map` function, so it only needs to be done once

nice, 👍 

> Clarify that ticks contain only label values

this one is not as clear to me. in fact, the nomenclature at large implies there's no general consensus on what to call things, unfortunately:

https://www.google.com/search?q=axes%2Baxis%2Blabels&tbm=isch

i admit that it often trips me up as well, and uPlot's internal naming conventions are not 100% consistent (yet). i think i'll rename the current `axis.label` to `axis.title` to reduce some confusion. another point to consider it that `series.values` returns an object with "labels" as keys. you can see that in the time-periods demo with table-style legends: https://leeoniya.github.io/uPlot/demos/time-periods.html

in uPlot-land at the moment "label(s)" and "title" are not transformed from the data, while "value" and "values" are. at least that was the naming logic.

"tick values" would be better here (if that helps). and please split the commits up.

leeoniya:
eh, but then "`values` is an array of formatted tick labels" doesnt make sense :(

benmccann:
I dropped that from the PR. A lot of the reason it wasn't clear to me is that ["tick" generally means something different in Chart.js](https://github.com/chartjs/Chart.js/issues/4186#issuecomment-297691904) (though we do sometimes say "tick" when referring to the tick + label as there is one label per tick. We've been trying to cleanup some of that inconsistency, but some probably still remains).

I feel like "value(s)" is a bit vague and possibly opposite of what I'd expect. I'd expect the tick value to be the timestamp and the label to be the date string

I do like the idea of renaming the axis label to title and think it's a good idea we might have to steal for Chart.js too as we've had the same confusion over having multiple things called label.

leeoniya:
> A lot of the reason it wasn't clear to me is that "tick" generally means something different in Chart.js

- text highlighted in red is the tick label
- purple box is the tick mark

i think this terminology is pretty unambiguous, they're not "axis labels". IMO "axis title" would be rather unorthodox and i'm not particularly eager to adopt it.

for me, the biggest issue with referring tick labels as "labels" is that a series "label" is static text. i don't want to have an axis "tickLabels" be a formatter but series "label" be static and have series' "value" be the formatter in that case. as it is now, everything called "value" in the API is a formatter and everything called label (or title) is static text. it's a different kind of consistency, but one that's easy enough to explain and does make some sense:

- a series label is static and the series value is formatted from the data
- an axis label is static and axis values are formatted from the raw tick data (which are derived from the scales but have the same datatype and semantic meaning as if they were from the data itself).

for Chart.js, your terminology has to be more general since you support a lot of different chart types. also, your API doesn't shy away from using unambiguous [albeit wordy] terms.

i'll keep it the way it is and just have to explain it better. naming things is hard 🤷‍♂

benmccann:
> they're not "axis labels". IMO "axis title" would be rather unorthodox and i'm not particularly eager to adopt it.

To be clear, the "axis label" is something different in Chart.js than either the red or purple box. Here's an image where I've circled the "axis label" in orange.

chartjs-terms

The terms "raw value" and "formatted value" might be more clear than "label" and "value" (though a bit more verbose, but "rawValue" / "raw value" are used 17 times in the readme right now, so it's inline with what's currently being done).  Using "label" and "value" to refer to formatting is rather unconventional and I imagine even if it's explained upfront a lot of people will miss that.

leeoniya:
this is where a picture is worth 1e3 words. uPlot's uniformity is in both naming, form & function:

- series label / series value
- axis label / axis values

if i called the axis values anything other than "values", it would catering to one expectation while breaking API uniformity. i do understand the pre-existing convention of calling them tick labels, but i'll sleep better if the API doesn't conflate the two for the sake of what a [non-RTFM] user expects.

i plan to include this image in the docs:

labels-vs-values

PR #85, opened 2020-01-03
Fix error in Chart.js-next benchmark

benmccann:
The way to disable animations has changed


leeoniya:
i suspect more changes are coming to this impl in https://github.com/chartjs/Chart.js/pull/6923. i'd prefer those changes (and any others that affect perf) get rolled into this PR rather than making new ones each time `next` gets updates.

Issue #87, opened 2020-01-04
implement API primitives to support dual chart zoom ranging


leeoniya:
copied from https://github.com/leeoniya/uPlot/issues/83#issuecomment-570573947

i was planning on making a demo similar to this:

http://dygraphs.com/gallery/#g/range-selector

there's nothing built in but i think there's enough api to make this without too much effort using 2 charts with the same data but different scale ranges and some ui sprinkled on.

i might need to add the ability to disable zoom without disabling selection and make a new "select" hook. i havent really thought about it much yet.

leeoniya:
most of the necessary work is done in 10bf5b2b5739c41b1b3aeec4055ba21510f0a83b & 130fc217a2d29b1f6903951f78b1b036a6a01c49. i have a basic bi-directional demo working locally, but still not completely happy with the hook firing order and *when* setSelect should fire:

- before or after `init`? probably after since setCursor fires after (as a result of setData)
- should it fire when the selection is set *and* cleared? (right now i'm not firing it when it's cleared)
- should it fire on every setCursor or just mouseup (probably the latter).

still need to add the ability for initial (but not permanent) scale ranges to be be defined.

the bi-directional nature leads to a chicken-egg problem if we rely only on hooks to set initial state. does the zoomed chart's setScale trigger the full chart's setSelect or vice versa? if neither and we rely on setting the values directly via opts, does that mean we don't fire the hooks at all? bad too.

mbj4668:
Can you share the demo you have?

leeoniya:
yes, i'll push it later tonight.

leeoniya:
the demo is here: https://github.com/leeoniya/uPlot/blob/master/demos/zoom-ranger.html

not super elegant, but not terrible either. couple thoughts that came out of this exercise.

- `.setCursor()`, `.setSelect()` (and their matched initial opts values `.cursor` and `.select`) take left/width/top/height coords, but this is mostly useless from an API perspective because 1) it would be much better to set the cursor or selection to some specific x and y scale values, or values from the data and 2) during init, there is no way to know what coords to specify before the chart & data if fully initialized with the scales and we cannot yet make use of `.valToPos()`. i think these apis will change to accept scale values rather than coords.

[...]

PR #89, opened 2020-01-06
gitignore + beginDblClick hook

danyalejandro:
This adds a .gitignore generic file, removes an editor file, and adds a dblClick hook that allows to do stuff (i.e. switch the data) before dblClick does its thing.


danyalejandro:
My current use case: I want to change the plot data on dblClick zoom-out. dblClick calculates the ranges from the data before calling setScale(), so it's not straightforward how to do this without the hook.
Also, hooks rock.

leeoniya:
> This adds a .gitignore generic file, removes an editor file

`.editorconfig` is not there by accident. it's there so that github adds `.tab-size[data-tab-size="4"]` to source views in this repo. otherwise the default tabsize is 8. https://stackoverflow.com/a/33831598. nothing in `.gitignore` applies to my environment. if you'd like to exclude those things locally, simply add them to your local `.git/info/exclude` file.

> My current use case: I want to change the plot data on dblClick zoom-out. dblClick calculates the ranges from the data before calling setScale(), so it's not straightforward how to do this without the hook.

seems like a legit requirement. however, i've intentionally stayed away from naming and using hooks as a proxy for event handlers. if anything like this gets added, it should get added generically as a `willSetScale` hook or similar. but then we'd want a will/did pair for all hooks and not just some. i'm not terribly eager to do this, though [it's been discussed](https://github.com/leeoniya/uPlot/pull/76#issuecomment-568540095).

let me think about this for a bit.

danyalejandro:
Makes sense, sorry for the confusion.

On your comment though, I don't know if `willSetScale` would solve this for me, because of this:

`function dblClick(e, src, _x, _y, _w, _h, _i) {
		//fire("beginDblClick");
		let min = data[0][0];
		let max = data[0][dataLen - 1];
		_setScale(xScaleKey, min, max);
`
It's already too late to change the data when `_setScale` runs because min & max got calculated from the old data.

I also thought about `begin/will` & `end/did` hooks for all mouse events, but dblClick() seems to be the only function that requires a hook; everything else in my use case can be done via `setScale` as you described.

leeoniya:
there's another place where the scale can be adjusted: `scale.range`. it accepts the initial/intended min/max should return the final min/max. its actual purpose is to provide an override/adjust any default scale min/max snapping.

https://jsfiddle.net/32p5h016/

this is probably not the ideal place to do this, certainly feels pretty hackish (and certainly not 100% done).

danyalejandro:
Yeah, your example certainly works, but since the range is calculated from dynamic data I'm not sure `scales.x.range` would be the obvious place to keep this logic (unless you've gone through the source code).

leeoniya:
there's actually a lot in common here with #87.

as of recently, you can pretty easily disable and re-implement zooming in userland without relying on `setScale`, but on the new `setSelect`. i think this will be the way to go:

https://jsfiddle.net/5jL43bty/5/

the only part that's missing is the ability to disable the native dblclick handler, but i think i'll just assume this if you're already setting `cursor.drag.setScale: false`.

you can see that the setScale call precedes setData, because passing `false` to setData does not actually prevent `setScale`, but prevents resetting it to the full data range. with `false`, it re-uses the current scale's min-max, but still redraws. maybe setData/false should prevent any kind of setScale invocation.

i'm currently in the process of ironing out some more hook/init weirdness for #87, and will probably prevent `setSelect` and `setCursor` hooks from firing during init (they're both hidden initially anyhow). this would allow to drop `if (!initial) {` from the demo.

stay tuned...

leeoniya:
ok, this is now good: https://github.com/leeoniya/uPlot/blob/master/demos/fetch-zoom.html

danyalejandro:
Took me a few days to implement setSelect but I finally managed to make it work (wheel zoom included). Thank you!

leeoniya:
> Took me a few days to implement setSelect

sounds like a while, any specific difficulties?

have you seen/used https://leeoniya.github.io/uPlot/demos/wheel-zoom.html as reference?

danyalejandro:
Your wheel-zoom and fetch-zoom examples got me started, thanks a lot for both.

I guess it's a bit the lack of documentation on hooks (I reload data a lot). I know uPlot is still in development so docs will come in due time. My timestamps are in milliseconds too, so I have to go back and forth between s and ms.

It's just my use case.

PR #90, opened 2020-01-07
Shave a couple milliseconds off Chart.js benchmark

benmccann:


leeoniya:
thanks, next re-bench will be after Chrome 80.

benmccann:
No need to rebenchmark. While it's faster, it's probably not by a measurable amount

Issue #80, opened 2019-12-31
Feature suggestion: redraw()


ldstein:
Hi @leeoniya 

I'd like to suggest adding a 'redraw()` function on the public API. The purpose would be to allow canvas redraws (ie `paint()`), even when the internal chart state has not changed.

Use Case: I'm highlighting regions of a chart using the `drawClear` hook. Regions can be added/removed/resized by the user, which in turn requires the chart to be redrawn. 

Currently there doesn't appear to be a straightforward way to programmatically trigger a redraw without changing the data or scale (though please correct me if I'm wrong). 

My current workaround is to simply destroy / initialize a new chart every time my data changes.

leeoniya:
was easy enough ;)

leeoniya:
btw, i was thinking of splitting up the draw-hooks demo into several applied plugins rather than having one monolithic one (but still keep everything in one file). would you be interested in doing it? 

i was thinking something along these lines:

- `bgGradientPlugin({stops: [...]})`
- `gridBlurPlugin({radius: 2})`
- `seriesMediansPlugin(opts)`
- `seriesPointsPlugin({spikes: 6})`
- `renderStatsPlugin(opts)`

ldstein:
Thanks for that. Re demos, no problem. Will see if I can get it together next weekend.

leeoniya:
ping @ldstein

ldstein:
Sorry! Completely fell off my radar. See #96

PR #96, opened 2020-01-22
Refactor draw hooks demo

ldstein:
Refactor of demos/draw-hooks.html as per discussion in #80


leeoniya:
hey, thanks. i'm not sure this was the desired result:

draw-hooks

ldstein:
Yeah, that was definitely not the desired result. PR updated.

Issue #36, opened 2019-10-28
some kind resize(width, height) API


danyalejandro:
Use case: Our sensor dashboards are shared by a team of 7 scientists. Some of them have their screens rotated (portrait). They both want to maximize screen usage without changing the code.

Would it be possible to add a "**stretch to fit**" option?

![image](https://user-images.githubusercontent.com/1218608/67707638-1417b300-f978-11e9-94e6-7769fe9cac8a.png)

![image](https://user-images.githubusercontent.com/1218608/67707805-63f67a00-f978-11e9-966c-350b0dccd06f.png)

leeoniya:
i would recommend that you first measure the size of the container you intend to put the charts into using something like [element.getBoundingClientRect()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) and use those metrics to create the uPlot instances.

i have no plans to implement auto or fluid sizing in the core, since most of this can be done handled outside the lib. and most of the time is not necessary after initial render anyhow.

if you need to resize the plots after initial render. i can maybe add a `.resize(width, height)` method so you can use that after init.

danyalejandro:
A resize() would work perfectly!

leeoniya:
this is now available as `.setSize({width, height})`

mfiot-rm:
So I have an event listener and at certain break points I run:
```
size = document.getElementById("chart-wrap").getBoundingClientRect()
uplotD.setSize(size.width,size.height)
```
But the plot sorta disappears whenever the size is update.
![image](https://user-images.githubusercontent.com/78705290/132061945-8eba8f08-08a9-4673-a488-f024c9869d06.png)

Am I using the setSize wrong does it need some sort of redraw or update after?

leeoniya:
> Am I using the setSize wrong

yes, it takes a single `{width: ..., height: ...}` object, not two arguments.

PR #97, opened 2020-01-23
Added resize fn

ldstein:
Hi @leeoniya 

This is a resize function I've been tinkering with in the background. Demo available in `demos/resize.html`. Addresses  #36 

Some duplicate code between Chart initialization and calls to resize (Chart initialization could probably call `updateChartDimensions()` but wanted to gauge your interest first - wasn't sure if you've already implemented this.

### DIFF
--- a/demos/resize.html
+++ b/demos/resize.html
@@ -1,0 +1,112 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8">
+		<title>Draw Hooks</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1">
+
+        <link rel="stylesheet" href="../src/uPlot.css">
+        <style type="text/css">
+            #resizeForm {
+                background:red;
+                position:fixed;
+                bottom:10px;
+                left: 10px;
+
+            }
+
+            #chartWrapper {
+                position: fixed;
+                top:0;
+                right:0;
+                bottom:30px;
+                left:0;          
+                border:5px solid #ccc;      
+                box-sizing: border-box;
+                overflow: hidden;
+            }
+
+            .uplot .legend {
+                position: fixed;
+                left:0;
+                right:0;                
+                bottom:0;
+                height:30px;
+                background:#f3f3f3;
+            }
+
+        </style>
+	</head>
+	<body>
+        <div id="chartWrapper"></div>
+		<script src="../dist/uPlot.iife.js"></script>
+		<script>
+            "use strict";
+
+			function makeChart(domContainer) {
+				const data = [
+					[ 1, 2, 3, 4, 5, 6, 7, 9,10],
+					[0,1,8,16,20,25,18,19,6],
+				];
+
+				const grid = {
+					show: true,
+					color: '#000',
+					width: 1,
+				};
+
+				let opts = {
+					width: domContainer.clientWidth,
+					height: domContainer.clientHeight,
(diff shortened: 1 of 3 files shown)


leeoniya:
thanks, this is a good start.

i was holding off until the internals were mostly settled so that i didn't have to update more code than necessary during refactorings. at this point, i don't foresee any architectural overhauls, so this is welcome.

i think we should stick with the current API conventions and call it `.setSize({width, height})` and also add a matching hook for it.

i'm not sure i'm on board with the internal function naming, but i can fix that as a follow-up myself - i'll give those some thought.

as for the demo itself, i don't know if it would feel nicer with a 100ms throttle rather than a 100ms debounce.

> Chart initialization could probably call `updateChartDimensions()`

yeah, probably.

demos/resize.html
```diff
@@ -0,0 +1,112 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8">
+		<title>Draw Hooks</title>
```
leeoniya: title

demos/resize.html
```diff
@@ -0,0 +1,112 @@
...
+				return new uPlot.Line(opts, data, domContainer);
+			}
+
+            let resizeTimeoutId = null;
+
+            function handleWindowResize(evt)
+            {
+                clearTimeout(resizeTimeoutId);
+
+                resizeTimeoutId = setTimeout(() => {
```
leeoniya: this callback can be hoisted to it's not recreated on every resize event.

demos/resize.html
```diff
@@ -0,0 +1,112 @@
...
+                    let width = chartWrapperEl.clientWidth;
+                    let height = chartWrapperEl.clientHeight;
+                    chartInstance.resize(width, height);
+
+                    chartInstance.setScale('x', chartInstance.scales.x.min, chartInstance.scales.x.max);
+                    chartInstance.setScale('y', chartInstance.scales.y.min, chartInstance.scales.y.max);
```
leeoniya: these two setScale call should be batched together via `u.batch()`.

src/Line.js
```diff
@@ -410,6 +435,43 @@ export function Line(opts, data, then) {
 	self.ctx = ctx;
 
 	plot.appendChild(can);
+	
+	self.resize = updateChartDimensions;
+
+	function resize(newWidth, newHeight) {
+		updateChartDimensions(newWidth, newHeight);
+		Object.keys(scales).forEach(id => setScale(id, scales[id].min, scales[id].max));
```
leeoniya: the rest of the codebase uses `for-in` iteration over `scales` to avoid array allocation that `Object.keys` does

demos/resize.html
leeoniya: thinking about it more, what's a scenario when .resize() would be called and the user would not want all the scales auto-updated? it's quite the burden to put on users to expect to re-sync all the scales and also know to batch it all. my expectation would be that the scales auto-update - that's the most common case imo.

leeoniya:
hey, thanks for this PR, i decided to work on this myself (the changes are large enough where it'll take a lot of back-and-forth to get this PR to where i want).

should have some commits soon!

leeoniya:
this is now available as `.setSize({width, height})`.

ldstein:
Awesome, thanks @leeoniya

Issue #119, opened 2020-02-05
Highlight intervals (weekends, etc..)


CrashLaker:
Hi all,

Is there a way to highlight intervals along the X axis? e.g. weekends, holidays, etc.
Like in:
http://dygraphs.com/gallery/#g/highlighted-weekends

leeoniya:
nothing out of the box, but it should be easy enough to author a plugin that does this. you have full access to all timestamps obviously, and you can use `uPlot.tzDate()` to get a timezone-adjusted date objects which you can use to isolate weekend ranges, then simply paint on `u.ctx.canvas` in the `drawGrid` hook if want it atop the grid (but below the series), or `drawClear` hook if you need it under the grid.

take a look at how https://github.com/leeoniya/uPlot/blob/master/demos/draw-hooks.html does everything.

would be a good plugin to author for the demos here ;)

leeoniya:
copying from https://github.com/leeoniya/uPlot/pull/126#issuecomment-594074971:

> yes, you have to use `uPlot.tzDate`. the real trick is not to iterate every datapoint in the scale range in search of weekends, because that would mean creating 60,000 Date instances when you have 30,000 datapoints. you have to find these weekends efficiently, like the axis ticks find breaks in months, days, hours without walking the data.

PR #126, opened 2020-02-13
Added highlight weekends demo

ldstein:
This PR demonstrates a very basic highlight weekends plugin. 

It (or something similar) could be used in #119 .

image

### DIFF
--- a/demos/highlight-weekends.html
+++ b/demos/highlight-weekends.html
@@ -1,0 +1,94 @@
+<!doctype html>
+<html>
+
+<head>
+	<meta charset="utf-8">
+	<title>Highlight Weekends</title>
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<link rel="stylesheet" href="../src/uPlot.css">
+</head>
+
+<body>
+	<script src="../dist/uPlot.iife.js"></script>
+	<script>
+		function highlightWeekendsPlugin(fillStyle = 'rgba(255,255,204') {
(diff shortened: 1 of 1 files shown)


leeoniya:
thanks @ldstein 

as you can see from your image, `new Date` fails to align to actual day boundaries. this simplr version will only work when both your machine and the desired timezone in which you want to view the data are the same. a proper solution would have to use `uPlot.tzDate` and replicate some of the same logic used internally to place axis ticks. this makes writing plugins along the time scale particularly nuanced/challenging.

i feel like it's worth exposing some sort of `ticks()` method which can reusue the logic and return temporal breakpoints at a higher resolution than the ticks generated for the grid. this can then be used to help with situations like this. one limitation is that the times cannot be arbitrary, but have to align to logical markers.

ldstein:
Thanks @leeoniya . Clearly a more complicated issue than I anticipated.

I've closed the PR. Will reopen once I'm able to revisit.

CrashLaker:
Hi.

As I understood from this we need to use the `tzDate` passed in `opts` otherwise the `new Date()` approach would work just fine right?

Atm can we access the `opts` inside the plugin's scope?

leeoniya:
yes, you have to use `uPlot.tzDate`. the real trick is not to iterate every datapoint in the scale range in search of weekends, because that would mean creating 60,000 Date instances when you have 30,000 datapoints. you have to find these weekends efficiently, like the axis ticks find breaks in months, days, hours without walking the data.

Issue #129, opened 2020-02-16
Zoom Endstop


GatCode:
Hello,
I am struggling to stop the diagram at the endpoints of the data.

For example when a user zooms out completely, a view like the one below is possible.
![Bildschirmfoto 2020-02-16 um 16 10 07](https://user-images.githubusercontent.com/19287308/74607207-20ff1480-50d7-11ea-8a64-4d84da6716a6.jpg)

I tried to alter the setScale option to stop at the endpoints but I couldn't achieve a smooth result.

In the end I would like that no matter how much a user zooms out, the maximum would be this:
![Bildschirmfoto 2020-02-16 um 16 10 19](https://user-images.githubusercontent.com/19287308/74607224-5146b300-50d7-11ea-9233-8942e634c622.jpg)

Has someone already fixed my problem or could someone please help me out?

leeoniya:
i'll probably make this the default behavior for the wheel-zoom demo. it should be just a matter of clamping `nxMin`, `nxMax` and `nxRange` after comparing to data[0] min/max values.

i'm currently working on #112, so won't get to this until that's resolved.

https://github.com/leeoniya/uPlot/blob/dd07c04deb504a2148ec11a81db10035ce1e98c8/demos/zoom-wheel.html#L31-L33

leeoniya:
this is done.

PR #132, opened 2020-02-24
Added zoom out bounds to zoom wheel demo

ldstein:
This PR adds a limit to how far you can zoom out in `zoom-wheel.html`, as discussed in #129.

### DIFF
--- a/demos/zoom-wheel.html
+++ b/demos/zoom-wheel.html
@@ -12,43 +12,39 @@
 		<script>
 			function wheelZoomPlugin(opts) {
 				let factor = opts.factor || 0.75;
-
-				function getDataBounds(data) {
-
-					const bounds = {
-						minX : null,
-						minY : null,
-						maxX : null,
-						maxY : null,
-					};
-
-					data.forEach((series, i) => {
-
-						let minVar = i == 0 ? "minX" : "minY";
-						let maxVar = i == 0 ? "maxX" : "maxY";
-
-						series.forEach((val) => {
-							bounds[minVar] = (bounds[minVar] === null) ? val : Math.min(bounds[minVar], val);
-							bounds[maxVar] = (bounds[maxVar] === null) ? val : Math.max(bounds[maxVar], val);
-						});
-					});
-
-					return bounds;
-				}
+				let bounds = null;
 
 				return {
 					hooks: {
+						setScale: (u) => {
+							// Use the same values uplot internally calculates for resetting scale
+							if (!bounds) {
+								bounds = {
+									minX: u.scales.x.min,
+									maxX: u.scales.x.max,
+									minY: u.scales.y.min,
+									maxY: u.scales.y.max
+								}
+							};
+						},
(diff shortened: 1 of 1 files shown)


leeoniya:
hey, thanks.

this is not quite right, though. try zooming in on one point at the right-hand side and then back out from a point on the left-hand side. it's also possible to "unzoom" from the initial fully-unzoomed view.

i think the way this should work if to honor the requested zoom range (maxX-minX) but shift it by the amount that keeps it within the data bounds. only when the maxX-minX is > data range, then we can dishonor the requested zoom range and clamp it to data bounds.

ldstein:
PR updated. 

Quick question: Why is the initial min/max y scale 16/82 when the data bounds is 18/80?

I noticed by adding `scales.y.range: (u, dataMin, dataMax) => [dataMin, dataMax])` will make the grid perfectly align to the data, but assume there is a reason behind it.

Note the x scale is perfectly matched to the x data bounds.

leeoniya:
the behavior looks good now. quite a bit more code though than expected. i'll review it in more detail soon.

> Quick question: Why is the initial y scale min/max calculated to 16/82 when the data bounds is 18/80?

typically, you want a buffer above and below the extremes:

https://www.google.com/search?q=line+chart&tbm=isch

there is also special casing for when the data never goes < 0. uPlot actually exposes its numeric padding logic as `uPlot.rangeNum()`. you can see it being re-used here:

https://github.com/leeoniya/uPlot/blob/a9f777f9abd3196a2b1806434cfaac2ce9d0ec33/demos/pad-scale.html#L27-L28

but getting a bit more into the weeds, the extra scale padding gives uPlot's tick gen an opportunity to place an extra tick above/below the data min/max. the current logic is:

1. get a scale's initial min/max by analyzing all series on that scale
2. add an "extra tick" buffer to scale min/max as determined by rangeNum()
3. calculate ticks for the scale's new min/max

this is imperfect because it does not really guarantee that we'll actually get another tick, but i think in most cases it works out. another downside to this approach is that you're not guaranteed to have ticks at the chart limits, which basically sucks.

ideally we would calculate needed ticks based on the the initial scale range, then use that information to expand the scale min/max to an exact tick multiple, and then recalculate the ticks again. however, the new ticks may completely change due to the minimum required tick spacing and our new "edge" ticks will no longer be correct and you end up in the same place as you started.

the rangeNum() logic should be improved to expand the scale in a way that somehow guarantees that the tick gen will create edge ticks.

if you want to ponder a solution to this, it would be a good improvement.

ldstein:
Great, thanks for detailed explanation.

leeoniya:
i took another blind crack at this myself, then went to review this PR and saw basically identical logic :)

i factored the clamping out into a helper, which cut the line count down by 50% and is more clear as a result. i also put everything into a `ready` hook, and fixate the max scale limits there. i shamelessly stole your `e.preventDefault();`

thanks for your help with this!

PR #135, opened 2020-02-26
Bugfix for issue #134 - axis.space is not a function

### LINKED ISSUE #134: axis.space is not a function
If the x-axis (axis 0) is hidden via `show: false` then the error `axis.space is not a function` is thrown from inside `getIncrSpace`.

switchtrue:
Fixes #134


leeoniya:
thanks!

PR #136, opened 2020-02-28
Chart.js no longer needs CSS

benmccann:
CSS is no longer used to detect chart resizes since https://github.com/chartjs/Chart.js/pull/7104 so the CSS file no longer exists


benmccann:
Btw, I wouldn't worry about re-benchmarking. I was just trying to keep the file working. We'll release an alpha at some point and then we can probably just switch this file to use the alpha instead of `master` so that it doesn't break whenever we change the code

PR #191, opened 2020-05-01
< 25 kb is very small

NotMoni:


leeoniya:
@NotMoni 

thanks for the PR, but i'm actually more inclined to go the other way and just let the numbers speak for themselves.

i consider https://en.wikipedia.org/wiki/.kkrieger to be very small, just for some perspective :D

cheers!

NotMoni:
Oh ok 👍

PR #193, opened 2020-05-02
fix definition according to source code

ALEUT:
1. `names` parameter for `fmtDate` is optional
2. `fmtDate` is optional
3. array -> Array (array can be defined in TS as either `Array` or `[]`).
4. definitions can't have `= value` for fields.


leeoniya:
awesome, thanks!

PR #197, opened 2020-05-05
Fix type definitions, fixes #196

### LINKED ISSUE #196: Errors in d.ts
Steps to reproduce:
1. Install uPlot 1.0.8 from npm.
2. In a TS 3.8.3 project with strict typing, `import uPlot from 'uplot'`.
3. Attempt to build the project.

What happens:
Numerous errors during compilation.
image

Expected behavior:
No errors.

Klaster1:


leeoniya:
thanks!

```ts
setData(data: uPlot.AlignedData, resetScales?: boolean): void;
```

it's really unfortunate that the arg defaults cannot be hinted. now i have to guess what the default behavior of `setData` is with regards to `resetScales`. the only simple solution is to add it to the doc comment, i assume?

Klaster1:
@leeoniya good point, I've added comments about default values, this preserves the original intent as much as possible.

leeoniya:
thanks, merged. i did a follow-up commit to flatten out the doc comments.

> You might want to consider a rewrite in TS

as you can see, i prefer terse code that's quick to scan. nothing against TS, but to me it's much harder to scan TS code with inlined types, etc. keeping the typedefs in sync is not so difficult in a project of this size to justify a codebase that's less optimal for me to work on.

PR #198, opened 2020-05-05
update demos/tooltips.html to work with custom scales

mezga0153:
tooltips.html assumed all series used 'y' as scale, this commit changes that to use series specific scale which should ease integrating tooltips in custom charts


leeoniya:
thanks, hard to argue with that!

merged, but for some reason Girhub didn't auto-close this PR :/

leeoniya:
ah, i merged the wrong branch from your repo. just re-merged your master and force-pushed.

PR #202, opened 2020-05-07
Fix type compilation

Klaster1:
Apparently, my previous PR #197 was not enough, after rebase to master more type compilation errors have surfaced:
image
What was done:
1. Added SpanGap type, inferred from "uPlot.js".
2. Used the new type to fix TS error from the screenshot above.

### DIFF
--- a/dist/uPlot.esm.d.ts
+++ b/dist/uPlot.esm.d.ts
@@ -276,7 +276,7 @@ declare namespace uPlot {
 		max?: number,
 	}
 
-	export type SpanGap = [number, number]
+	export type DataGap = [number, number]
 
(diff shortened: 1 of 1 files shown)


dist/uPlot.esm.d.ts
```diff
@@ -276,6 +276,8 @@ declare namespace uPlot {
 		max?: number,
 	}
 
+	export type SpanGap = [number, number]
```
leeoniya, line 279: let's make this `export type DataGap = [number, number];`

Klaster1: Done.

Issue #215, opened 2020-05-11
Sync cursor with different time axes


EmiPhil:
Looking for a way to combine the brush (ranger) demo and the sync cursor demo so that when you highlight your cursor on one it is properly placed in the other.

Is this doable out of the box? If not, what hooks would you suggest using to achieve it?

In my use case the time stamps of the "bigger" chart should be a subset of those in the ranger if that helps at all.

Thanks!

EmiPhil:
Adding a screenshot for more context in case it isn't super clear. The goal is for each of the Time values to be the same (or, if hovering in the ranger, then no time would be visible on the main chart if the ranger time is out of bounds of the main chart).

![Screenshot from 2020-05-11 15-01-48](https://user-images.githubusercontent.com/11808725/81612329-e5a07900-9399-11ea-8f17-ebcaebf1d899.png)

EmiPhil:
So I was able to get this working (mostly, still some kinks...note that the times aren't quite the same). I can post code tomorrow because I'd still be interested in your thoughts on methodology. I had to stop using the sync feature to get it to work (probably rerolled a more specific version of it). For now here is an updated screenshot.

![Screenshot from 2020-05-11 17-04-44](https://user-images.githubusercontent.com/11808725/81620773-e8a36580-93a9-11ea-83f0-c63b5e9ed889.png)

leeoniya:
hmm, interesting.

re-looking at the mousemove sync code (below), it uses the coordinate % along both x and y, so yeah you likely won't be able to use the native sync since it assumes matched x ranges.

we can prototype adding some option to `cursor.sync` to instead use the scale's value at the cursor position. the existing code already passes along the originating uplot instance, so it should be pretty easy to get this working. i guess we'd need to offer an x-scale and y-scale key to sync by:

`cursor.sync.scales: [xKey, yKey]`.

interested giving it the old college try?

https://github.com/leeoniya/uPlot/blob/b5fefa47977e79dd513009008e0ef62fc51411a3/src/uPlot.js#L1664-L1704

PR #216, opened 2020-05-12
Cursor scale value sync

EmiPhil:
Adds a scales array option to the cursor.sync of the signature [xKey, yKey] (default [null, null]) and includes a new demo showcasing functionality.

#215

### DIFF
--- a/dist/uPlot.d.ts
+++ b/dist/uPlot.d.ts
@@ -115,6 +115,8 @@ declare class uPlot {
 declare namespace uPlot {
 	export type AlignedData = readonly (number | null)[][];
 
+	export type SyncScales = [string, string];
+
 	export type MinMax = [number, number];
 
 	export interface DateNames {
@@ -245,6 +247,8 @@ declare namespace uPlot {
 			key: string;
 			/** determines if series toggling and focus via cursor is synced across charts */
 			setSeries?: boolean; // true
+			/** sets the x and y scales to sync by values. null will sync by relative (%) position */
+			scales?: SyncScales; // [xScaleKey, null]
 		};
 
 		/** focus series closest to cursor */
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -1343,7 +1343,7 @@ export default function uPlot(opts, data, then) {
 	const drag = FEAT_CURSOR && cursor.drag;
 
 	let dragX = FEAT_CURSOR && drag.x;
-	let dragY = FEAT_CURSOR && drag.y;;
+	let dragY = FEAT_CURSOR && drag.y;
 
 	if (FEAT_CURSOR && cursor.show) {
 		let c = "cursor-";
@@ -1477,8 +1477,14 @@ export default function uPlot(opts, data, then) {
 	}
 
 	function scaleValueAtPos(pos, scale) {
-		let dim = scale == xScaleKey ? plotWidCss : plotHgtCss;
-		let pct = clamp(pos / dim, 0, 1);
+		let dim = plotWidCss;
+		if (scale != xScaleKey) {
+			dim = plotHgtCss;
+			// invert the pos on the y axis
+			pos = dim - pos;
+		}
+
+		let pct = pos / dim;
 
 		let sc = scales[scale];
 		let d = sc.max - sc.min;
@@ -1492,7 +1498,7 @@ export default function uPlot(opts, data, then) {
 
 	self.valToIdx = val => closestIdx(val, data[0]);
 	self.posToIdx = closestIdxFromXpos;
-	self.posToVal = (pos, scale) => scaleValueAtPos(scale == xScaleKey ? pos : plotHgtCss - pos, scale);
+	self.posToVal = scaleValueAtPos;
 	self.valToPos = (val, scale, can) => (
 		scale == xScaleKey ?
 		getXPos(val, scales[scale],
@@ -1532,7 +1538,7 @@ export default function uPlot(opts, data, then) {
 
 	let cursorRaf = 0;
 
-	function updateCursor(ts) {
+	function updateCursor(ts, src) {
 		if (inBatch) {
 			shouldUpdateCursor = true;
 			return;
@@ -1612,51 +1618,86 @@ export default function uPlot(opts, data, then) {
 		}
 
 		// nit: cursor.drag.setSelect is assumed always true
-		if (mouseLeft1 >= 0 && select.show && dragging) {
-			// setSelect should not be triggered on move events
+		if (select.show && dragging) {
+			if (src != null) {
+				let [xKey, yKey] = syncOpts.scales;
 
-			dragX = drag.x;
-			dragY = drag.y;
+				if (xKey) {
+					let sc = scales[xKey];
+					let srcLeft = src.posToVal(src.select[LEFT], xKey);
+					let srcRight = src.posToVal(src.select[LEFT] + src.select[WIDTH], xKey);
 
-			let uni = drag.uni;
+					select[LEFT] = getXPos(srcLeft, sc, plotWidCss, 0);
+					select[WIDTH] = abs(select[LEFT] - getXPos(srcRight, sc, plotWidCss, 0));
 
-			if (uni != null) {
-				let dx = abs(mouseLeft0 - mouseLeft1);
-				let dy = abs(mouseTop0 - mouseTop1);
+					setStylePx(selectDiv, LEFT, select[LEFT]);
+					setStylePx(selectDiv, WIDTH, select[WIDTH]);
 
-				dragX = dx >= uni;
-				dragY = dy >= uni;
-
-				// force unidirectionality when both are under uni limit
-				if (!dragX && !dragY) {
-					if (dy > dx)
-						dragY = true;
-					else
-						dragX = true;
+					if (!yKey) {
+						setStylePx(selectDiv, TOP, select[TOP] = 0);
+						setStylePx(selectDiv, HEIGHT, select[HEIGHT] = plotHgtCss);
+					}
 				}
-			}
 
-			if (dragX) {
-				let minX = min(mouseLeft0, mouseLeft1);
-				let maxX = max(mouseLeft0, mouseLeft1);
-				setStylePx(selectDiv, LEFT,  select[LEFT] = minX);
-				setStylePx(selectDiv, WIDTH, select[WIDTH] = maxX - minX);
+				if (yKey) {
+					let sc = scales[yKey];
+					let srcTop = src.posToVal(src.select[TOP], yKey);
+					let srcBottom = src.posToVal(src.select[TOP] + src.select[HEIGHT], yKey);
 
-				if (uni != null && !dragY) {
-					setStylePx(selectDiv, TOP, select[TOP] = 0);
-					setStylePx(selectDiv, HEIGHT, select[HEIGHT] = plotHgtCss);
+					select[TOP] = getYPos(srcTop, sc, plotHgtCss, 0);
+					select[HEIGHT] = abs(select[TOP] - getYPos(srcBottom, sc, plotHgtCss, 0));
+
+					setStylePx(selectDiv, TOP, select[TOP]);
+					setStylePx(selectDiv, HEIGHT, select[HEIGHT]);
+
+					if (!xKey) {
+						setStylePx(selectDiv, LEFT, select[LEFT] = 0);
+						setStylePx(selectDiv, WIDTH, select[WIDTH] = plotWidCss);
+					}
 				}
-			}
 
-			if (dragY) {
-				let minY = min(mouseTop0, mouseTop1);
-				let maxY = max(mouseTop0, mouseTop1);
-				setStylePx(selectDiv, TOP,    select[TOP] = minY);
-				setStylePx(selectDiv, HEIGHT, select[HEIGHT] = maxY - minY);
-
-				if (uni != null && !dragX) {
-					setStylePx(selectDiv, LEFT, select[LEFT] = 0);
-					setStylePx(selectDiv, WIDTH, select[WIDTH] = plotWidCss);
+			} else {
+				// setSelect should not be triggered on move events	
+				let uni = drag.uni;
+	
+				if (uni != null) {
+					let dx = abs(mouseLeft0 - mouseLeft1);
+					let dy = abs(mouseTop0 - mouseTop1);
+	
+					dragX = dx >= uni;
+					dragY = dy >= uni;
+	
+					// force unidirectionality when both are under uni limit
+					if (!dragX && !dragY) {
+						if (dy > dx)
+							dragY = true;
+						else
+							dragX = true;
+					}
+				}
+	
+				if (dragX) {
+					let minX = min(mouseLeft0, mouseLeft1);
+					let maxX = max(mouseLeft0, mouseLeft1);
+					setStylePx(selectDiv, LEFT,  select[LEFT] = minX);
+					setStylePx(selectDiv, WIDTH, select[WIDTH] = maxX - minX);
+	
+					if (!dragY) {
+						setStylePx(selectDiv, TOP, select[TOP] = 0);
+						setStylePx(selectDiv, HEIGHT, select[HEIGHT] = plotHgtCss);
+					}
+				}
+	
+				if (dragY) {
+					let minY = min(mouseTop0, mouseTop1);
+					let maxY = max(mouseTop0, mouseTop1);
+					setStylePx(selectDiv, TOP,    select[TOP] = minY);
+					setStylePx(selectDiv, HEIGHT, select[HEIGHT] = maxY - minY);
+	
+					if (!dragX) {
+						setStylePx(selectDiv, LEFT, select[LEFT] = 0);
+						setStylePx(selectDiv, WIDTH, select[WIDTH] = plotWidCss);
+					}
 				}
 			}
 		}
@@ -1707,17 +1748,32 @@ export default function uPlot(opts, data, then) {
 				cursorRaf = rAF(updateCursor);
 		}
 		else
-			updateCursor();
+			updateCursor(null, src);
 	}
 
 	function cacheMouse(e, src, _x, _y, _w, _h, _i, initial, snap) {
+		if (_x < 0 || _y < 0) {
+			mouseLeft1 = -10;
+			mouseTop1 = -10;
+			return;
+		}
+
 		if (e != null) {
 			_x = e.clientX - rect.left;
 			_y = e.clientY - rect.top;
 		}
 		else {
-			_x = plotWidCss * (_x/_w);
-			_y = plotHgtCss * (_y/_h);
+			let [xKey, yKey] = syncOpts.scales;
+
+			if (xKey != null)
+				_x = getXPos(src.posToVal(_x, xKey), scales[xKey], plotWidCss, 0);
+			else
+				_x = plotWidCss * (_x/_w);
+
+			if (yKey != null)
+				_y = getYPos(src.posToVal(_y, yKey), scales[yKey], plotHgtCss, 0);
+			else
+				_y = plotHgtCss * (_y/_h);
 		}
 
 		if (snap) {
@@ -1740,19 +1796,16 @@ export default function uPlot(opts, data, then) {
 
 	function hideSelect() {
 		setSelect({
-			width:	!drag.x ? plotWidCss : 0,
-			height:	!drag.y ? plotHgtCss : 0,
+			width: 0,
+			height: 0,
 		}, false);
 	}
 
 	function mouseDown(e, src, _x, _y, _w, _h, _i) {
-		if (e == null || filtMouse(e)) {
+		if (src != null || filtMouse(e)) {
 			dragging = true;
 
-			cacheMouse(e, src, _x, _y, _w, _h, _i, true, true);
-
-			if (select.show && (drag.x || drag.y))
-				hideSelect();
+			cacheMouse(e, src, _x, _y, _w, _h, _i, true, false);
 
 			if (e != null) {
 				on(mouseup, doc, mouseUp);
@@ -1762,51 +1815,54 @@ export default function uPlot(opts, data, then) {
 	}
 
 	function mouseUp(e, src, _x, _y, _w, _h, _i) {
-		if ((e == null || filtMouse(e))) {
+		if (src != null || filtMouse(e)) {
 			dragging = false;
 
 			cacheMouse(e, src, _x, _y, _w, _h, _i, false, true);
+			setSelect(select);
 
-			if (mouseLeft1 != mouseLeft0 || mouseTop1 != mouseTop0) {
-				setSelect(select);
+			if (drag.setScale && (select[WIDTH] || select[HEIGHT])) {
 
-				if (drag.setScale) {
-					batch(() => {
-						if (dragX) {
-							_setScale(xScaleKey,
-								scaleValueAtPos(select[LEFT], xScaleKey),
-								scaleValueAtPos(select[LEFT] + select[WIDTH], xScaleKey),
-							);
-						}
+				if (syncKey != null) {
+					dragX = drag.x;
+					dragY = drag.y;
+				}
 
-						if (dragY) {
-							for (let k in scales) {
-								let sc = scales[k];
+				batch(() => {
+					if (dragX) {
+						_setScale(xScaleKey,
+							scaleValueAtPos(select[LEFT], xScaleKey),
+							scaleValueAtPos(select[LEFT] + select[WIDTH], xScaleKey)
+						);
+					}
 
-								if (k != xScaleKey && sc.from == null) {
-									_setScale(k,
-										scaleValueAtPos(plotHgtCss - select[TOP] - select[HEIGHT], k),
-										scaleValueAtPos(plotHgtCss - select[TOP], k),
-									);
-								}
+					if (dragY) {
+						for (let k in scales) {
+							let sc = scales[k];
+
+							if (k != xScaleKey && sc.from == null) {
+								_setScale(k,
+									scaleValueAtPos(select[TOP] + select[HEIGHT], k),
+									scaleValueAtPos(select[TOP], k)
+								);
 							}
 						}
-					});
+					}
+				});
 
-					hideSelect();
-				}
+				hideSelect();
 			}
 			else if (cursor.lock) {
-				cursor.locked = !cursor.locked
+				cursor.locked = !cursor.locked;
 
 				if (!cursor.locked)
 					updateCursor();
 			}
+		}
 
-			if (e != null) {
-				off(mouseup, doc, mouseUp);
-				sync.pub(mouseup, self, mouseLeft1, mouseTop1, plotWidCss, plotHgtCss, null);
-			}
+		if (e != null) {
+			off(mouseup, doc, mouseUp);
+			sync.pub(mouseup, self, mouseLeft1, mouseTop1, plotWidCss, plotHgtCss, null);
 		}
 	}
 
@@ -1821,6 +1877,18 @@ export default function uPlot(opts, data, then) {
 
 	function dblClick(e, src, _x, _y, _w, _h, _i) {
 		autoScaleX();
+		
+		if (src != null && select.show && (drag.x || drag.y)) {
+			if (drag.setScale)
+				hideSelect();
+			else
+				setSelect({
+					[LEFT]: 0,
+					[WIDTH]: plotWidCss,
+					[TOP]: 0,
+					[HEIGHT]: plotHgtCss
+				});
+		}
 
 		if (e != null)
 			sync.pub(dblclick, self, mouseLeft1, mouseTop1, plotWidCss, plotHgtCss, null);
@@ -1873,6 +1941,7 @@ export default function uPlot(opts, data, then) {
 	const syncOpts = FEAT_CURSOR && assign({
 		key: null,
 		setSeries: false,
+		scales: [xScaleKey, null]
 	}, cursor.sync);
 
 	const syncKey = FEAT_CURSOR && syncOpts.key;
--- a/demos/index.html
+++ b/demos/index.html
@@ -70,7 +70,8 @@
 	<h2>Zooming</h2>
 	<a href="zoom-variations.html">Different zoom variants (adaptive, uni/omnidirectional)</a>
 	<a href="zoom-fetch.html">Fetch &amp; update data on zoom</a>
-	<a href="zoom-ranger.html">Secondary sync'd overview chart for zoom ranging</a>
+	<a href="zoom-ranger.html">Secondary sync'd overview chart for x-axis zoom ranging</a>
+	<a href="zoom-ranger-xy.html">Secondary sync'd overview chart for x/y-axis zoom ranging</a>
 	<a href="zoom-touch.html">Pinch zooming/panning plugin</a>
 	<a href="zoom-wheel.html">Mouswheel zooming plugin</a>
 
--- a/demos/zoom-ranger-xy.html
+++ b/demos/zoom-ranger-xy.html
@@ -2,7 +2,7 @@
 <html>
 	<head>
 		<meta charset="utf-8">
-		<title>Zoom Ranger</title>
+		<title>Zoom Ranger XY</title>
 		<meta name="viewport" content="width=device-width, initial-scale=1">
 
 		<link rel="stylesheet" href="../dist/uPlot.min.css">
@@ -10,6 +10,10 @@
 			.uplot {
 				display: block;
 				width: 800px;
+			}
+
+			.uplot .select {
+				background: rgba(127,0,256,0.3);
 			}
 		</style>
 	</head>
@@ -38,17 +42,18 @@
 				width: 800,
 				height: 100,
 				cursor: {
-					y: false,
 					points: {
 						show: false,
 					},
 					drag: {
 						setScale: false,
 						x: true,
-						y: false,
+						y: true,
+						uni: 10
 					},
 					sync: {
-						key: "moo"
+						key: "moo",
+						scales: ["x", "y"]
 					}
 				},
 				legend: {
@@ -62,7 +67,7 @@
 				series: [
 					{},
 					{
-						stroke: "red",
+						stroke: "rgba(0,0,0,0.8)",
 					}
 				],
 				hooks: {
@@ -86,10 +91,12 @@
 				cursor: {
 					drag: {
 						x: true,
-						y: false
+						y: true,
+						uni: 40
 					},
 					sync: {
-						key: "moo"
+						key: "moo",
+						scales: ["x", "y"]
 					}
 				},
 				scales: {
@@ -105,7 +112,7 @@
 					},
 					{
 						label: "sin(x)",
-						stroke: "red",
+						stroke: "purple",
 					}
 				]
 			};


EmiPhil:
Hmm something funky is happening to the ranger selection period with this code. Selecting a range in the ranger always snaps to the end of the chart. I'll keep looking at it...

EmiPhil:
Okay so now horz/vert scaling works as expected.

Having both the vertical and horizontal ranger synced at the same time would require some more effort. I think that having vertical rangers is pretty niche (and having both is even nicher) though so I'd throw it on the back burner. Basically because the cursors are synced, it moves the horizontal ranger vertically when adjusting the vertical ranger and vice versa. So there would need to be some sort of logic somewhere to prevent that behavior on rangers (maybe an isRanger flag or something to that effect could help).

leeoniya:
thanks @EmiPhil 

i'm a bit busy tomorrow but i'll review this in the next day or two. in the mean time, please rebase, squash the commits and exclude the `/dist` files to reduce merge/review noise (i'll build them myself as a followup).

EmiPhil:
Should be ready for review. The bit inside of the cacheMouse function makes sense to me (perhaps there is a better way of calling `valToPos` and `posToVal` than using `self` and `src`? I'm not sure).

The part I'm less confident about is the scaling sync. I haven't played with the code of the library enough yet to even know if putting logic in `mouseup` function makes the most sense (not to mention if it is missing edge cases etc).

EmiPhil:
I'll loop back to this issue once we finish off #219 because getting the scaling to work will become the next issue otherwise

leeoniya:
agreed. i'd want to land that first since it's more straightforward.

src/uPlot.js
```diff
@@ -1681,8 +1681,17 @@ export default function uPlot(opts, data, then) {
 			_y = e.clientY - rect.top;
 		}
 		else {
-			_x = plotWidCss * (_x/_w);
-			_y = plotHgtCss * (_y/_h);
+			// check for value based syncs in the src plot
+			let [xKey, yKey] = src.cursor.sync.scales;
```
leeoniya, line 1685: unless there's a reason why these config opts would differ between publisher (src) and subscriber (self), let's just pull this off of the `syncOpts`.

src/uPlot.js
```diff
@@ -1681,8 +1681,17 @@ export default function uPlot(opts, data, then) {
...
+			if (xKey != null)
+				_x = self.valToPos(src.posToVal(_x, xKey), xKey);
```
leeoniya, line 1687: we can definitely bypass `self.valToPos()` since we're not gonna need canvas pixels, let's call the lower-level `getXPos()` and `getYPos()` directly. `src.posToVal()` will probably stay (it's a bit cheaper).

src/uPlot.js
```diff
@@ -1737,22 +1746,37 @@ export default function uPlot(opts, data, then) {
 
 				if (drag.setScale) {
 					batch(() => {
-						if (drag.x) {
-							_setScale(xScaleKey,
-								scaleValueAtPos(select[LEFT], xScaleKey),
-								scaleValueAtPos(select[LEFT] + select[WIDTH], xScaleKey),
-							);
+						if (drag.x && !(src != null && !src.cursor.drag[xScaleKey])) {
```
leeoniya, line 1749: similar to comment above, unless src and self have different opts, we can just use the local `drag[xScaleKey]`. also, not loving the double-negatives here.

src/uPlot.js
```diff
@@ -1737,22 +1746,37 @@ export default function uPlot(opts, data, then) {
...
 								let sc = scales[k];
 
 								if (k != xScaleKey && sc.from == null) {
-									_setScale(k,
-										scaleValueAtPos(plotHgtCss - select[TOP] - select[HEIGHT], k),
-										scaleValueAtPos(plotHgtCss - select[TOP], k),
-									);
+									// we are changing scale based on a sync request which may not
+									// have the same scale that we do
+									if (src != null && !src.cursor.drag[k])
```
leeoniya, line 1769: local `drag[k]`?

dist/uPlot.esm.d.ts
```diff
@@ -115,6 +115,8 @@ declare class uPlot {
 declare namespace uPlot {
 	export type AlignedData = readonly (number | null)[][];
 
+	export type CursorScales = [string, string];
```
leeoniya, line 118: please rename to SyncScales

leeoniya:
left some comments. most related to the likelyhood of sync'd plots having different `scales` opts. i think we can assume they will all match, so we can switch to using local vars instead of stuff off of `src`.

might have missed something, but that's round 1.

also, please rebase. commit message should be "allow cursor & drag sync using scales instead of coords. close #215."

src/uPlot.js
leeoniya: it looks like `src != null` is serving the same purpose as `e == null` at the top of `mouseUp()` (to check if this is a synced event or a local event). in that case we should keep it uniform. i'd be okay switching the `e == null` condition since `src != null` may be a bit clearer.

EmiPhil:
Thanks, I should be able to get to this tomorrow morning.

In terms of behaviour, right now what I was trying to achieve with this pull is that if you specify the scale keys it will use the values of the data at that point (maybe we can call this absolute position or something) and if they are null it uses the % method (maybe we can call this relative position?)

Does that API make sense for you?

EmiPhil:
Oh I should also mention that I think we may need to be able to sync the x and y separately. I don't want to get too into the details in just a comment from my phone right now so I'll write up something more comprehensive about what I mean later.

leeoniya:
> Does that API make sense for you?

i'm confused. we had a whole discussion about it in #215, i understand both the intent and API :)

> Oh I should also mention that I think we may need to be able to sync the x and y separately.

are they not already synced separately?

leeoniya:
@EmiPhil let me know if you think we can wrap this up in the next couple days since i'd like to cut a new version. otherwise we'll push it to the next release cycle. thanks!

EmiPhil:
I won't guarantee it! I just finished up some work stuff and will be working on this for the rest of the afternoon. As long as there aren't any surprises I hope to finish soon

EmiPhil:
So this is much more cleaned up now.

When I talk about "sync x and y separately" I'm thinking ahead on how to make it work with both rangers at the same time. It's really tough to put into words so I can try to throw together an example at some point. And there's definitely a possibility that what I'm thinking about will be trivial to do! But that is for tomorrow to look at.

demos/sync-cursor-ranger.html
```diff
@@ -0,0 +1,152 @@
...
+						stroke: color,
+					}
+				],
+				hooks: {
+					ready: [
+						uRanger => {
+							if (vert) {
+								let top = uRanger.valToPos(initYmax, 'y');
+								let height = uRanger.valToPos(initYmin, 'y') - top;
+								let width = uRanger.root.querySelector(".over").getBoundingClientRect().width;
```
leeoniya, line 94: `uRanger.bbox.width / devicePixelRatio` should be the same (and more elegant)

demos/sync-cursor-ranger.html
```diff
@@ -0,0 +1,152 @@
...
+								uRanger.setSelect({ top, width, height }, false);
+							} else {
+								let left = Math.round(uRanger.valToPos(initXmin, 'x'));
+								let width = Math.round(uRanger.valToPos(initXmax, 'x')) - left;
+								let height = uRanger.root.querySelector(".over").getBoundingClientRect().height;
```
leeoniya, line 99: can pull this off of uRanger.bbox, too

src/uPlot.js
```diff
@@ -1485,14 +1485,18 @@ export default function uPlot(opts, data, then) {
 		return sc.min + pct * d;
 	}
 
+	function yScaleValueAtPos(pos, scale) {
+		return scaleValueAtPos(plotHgtCss - pos, scale);
```
leeoniya, line 1489: since `scaleValueAtPos()` already contains an internal check for `xScaleKey` to auto-set the `dim`. why not just move the automatic `pos` inversion into there as well, instead of making a dedicated `yScaleValueAtPos()`.

leeoniya:
> I'm thinking ahead on how to make it work with both rangers at the same time.

oh, i see. this might indeed be tricky, not to mention ultra-niche :D. i'm already somewhat hesitant to include the y-ranging demo for the same reason - i've never seen one or felt like i needed one. having the y ranger also makes the demo more complicated by adding an opts generator that uses `vert` conditionally :\

i'd like to leave any further sync complication out of the core. even if it's not a lot more code, every additional nuance makes maintenance more difficult, and the returns here are way past diminishing here, IMO.

EmiPhil:
Fair enough. If you don't want it in the core that makes this issue easier! If we take out the y axis example should we just put this example code into the ranger example that already exists?

EmiPhil:
Oh one thing that may not be as niche is wanting an x ranger but still wanting to be able to zoom on the y axis on the main chart. I'll play with that to see how it behaves (that's going to be the behavior I'm looking for, I agree the y ranger is weird, though it does really help to make sure the code is working!)

leeoniya:
> If we take out the y axis example should we just put this example code into the ranger example that already exists?

yeah i think that makes sense. i feel like the default behavior of sync should maybe change to `[xScaleKey, null]`. syncing x by position would still be useful for things like comparing different time periods across multiple charts, but that would be the less lesser case.

EmiPhil:
Okay will do. The latest push fixes up your last commend in uPlot.js and makes clicking behave a bit smarter.

EmiPhil:
Latest push adds an early return to `cacheMouse` if _x or _y are negative (it wasn't properly getting rid of the cursor lines on the other chart without it).

Also removes the x/y example and adds sync cursor to the ranger example (which lets us get rid of some of the plugin logic).

PS I would have liked to test the `sync-cursor.html` demo on my local comp but neither firefox or chrome allow me to load the data from the json file.

EmiPhil:
Hmm to get double clicking the main chart to work as expected I called `hideSelect` on synced charts that are still showing the select box. Actually I think the behavior would be opposite, on double clicking the main chart if the setScales drag property is false on sync charts set the ranger to max. I'll work on that.

EmiPhil:
Okay now we check for `drag.setScale`. If true we clear, if false we expand. This should be ready for (final?) review

leeoniya:
initial testing: `/demos/sync-cursor.html` no longer zooms properly

EmiPhil:
In the dst example (probably the same error as sync-cursor) the problem is that when `self` needs to set its scales it hides its select and then publishes the change (so now the sync charts receiving the request get a `src` where select has been cleared).

Easy fix is to publish first, do work after, but in most of the functions the workflow is work first publish after. I'll push the easy fix + html changes soon and you can comment

leeoniya:
hmm, mouseup is not really perf-sensitive, and it's quite odd if subscribers update before the publisher. i'd feel better if we cached the values temporarily and published with those, but still at the end.

EmiPhil:
Okay so now we store the select in a const at the start and reassign it at the end if we pub

leeoniya:
nice. looking a lot better.

i'd like to avoid `drag.setScale` affecting `setSelect` behavior, even if it means having to handle selection hiding in the demo via hooks. keeping them separate means that there's no magic interaction you need to account for when using the `setSelect` and `setScale` hooks.

leeoniya:
i'm still not seeing proper behavior in the other demos, tough. zooming does not match scales ranges in synced charts.

EmiPhil:
Huh. This is the weirdest thing. Console logging before and after:

`left (before) = 104`
`left (after) = 104.00000000003473`

I'll figure it out soon hopefully...

EmiPhil:
Oh I found the problem :facepalm:

EmiPhil:
It was insufficient to cache the select position because posToVal returns a value based on the current scale which changes if drag.setScale is true.

So, we have to cache the calculated values for all of the min and maxes before we set the scales. Then we can use those precalculated values directly if the update comes via src.

leeoniya:
thanks @EmiPhil 

getting this working properly is turning out to be quite a doozy, but i've made some more progress. your latest version unfortunately needs a lot of hacks to work, so here's what i did. i started with your previous version and re-worked the mouseUp's syncRequest to work more as a "push" rather than a "pull". this way, we simply compute the scales for the selection at the source and add them as an additional param to pub(). this avoids all the caching nonsense and becomes more straighforward.

take a look at this branch: https://github.com/leeoniya/uPlot/tree/cursor-scale-value-sync

it appears to be working for x, though the x + y seems to be buggy on double-click in the tweaked cursor-sync demo. this demo also doesnt technically have multiple (or differing) y scales in the charts, so that might be good to test. maybe you can resurrect your y-ranger and test it out. i'm off to get some sleep!

thanks!

EmiPhil:
Oh yay! The cache thing felt very hacky. I'll look at the code soon

EmiPhil:
For future reference I was finally able to get the sync-cursor example working by using the web server for chrome extension which mounts the local filesystem. Without it all browsers were blocking the demos from loading json data via `file:///` urls

leeoniya:
i usually just execute `http-server` in the repo dir: https://www.npmjs.com/package/http-server

leeoniya:
another thought i had is:

why do we need to do anything in mouseup at all if the cursor & selection sync already accounts for scale keys?

i think the issue is that cursor & select coords are clamped. if we could opt out of clamping when isSyncReq (or always), then we could probably remove all mouseUp changes (they still feel a bit like a hack). the cursor & select would still be visually clipped by the parent container (.over).

this might be another route to explore and i think is the actual solution here, honestly.

EmiPhil:
I'll look into that

EmiPhil:
Yeah so this api looks a lot better for sure!

I need to play more with the sync cursor example but I think this api fixes it. If you set the y key to null then it uses the relative position for all of them (which I think makes sense) but if you set the "y" key it behaves weird in a way that I think may be correct. Since two of the scales are using % they map correctly and the MB one maps weird but it's a bit like...of course it maps weird, those aren't the same numbers. I think. 

I have a meeting so I'll loop back to this in about 1-2 hrs

EmiPhil:
Okay!

From my perspective this behaves as you would expect it in most cases

I would recommend you play with the demos in this order:

`sync-cursor`
1. As is. MB is mapped to % because they are both on the y scale key.
2. Set the "y" scale sync option to null. Now they are mapped by relative position.
2. Optionally turn uni off, it works as expected.

`zoom-fetch`
1. As is. This is like the kitchen sink. The selection box is properly scaled between the charts and ticks at the uni points at the same time on both. This was also true for the difference sized charts in `sync-cursor` but it is a lot more noticeable here. The selection box of the ranger is also shows the full zoom of the main chart. I made it purple so you could really see what is going on with it.

I'm aware that most rangers don't do that and I had some ideas for only doing it when you are trying to sync x and y, but the more I played with it the more it just felt right to do it that way. Like if you don't do it that way and you sync the cursor to y then you get this (to me) odd behavior where the select box of the ranger is taller than you can ever get the cursor.

For the behavior most people are used to with a time ranger you need to set the ranger opts to:
```js
				cursor: {
					y: false,
					points: {
						show: false
					},
					drag: {
						setScale: false,
						x: true
					},
					sync: {
						key: "moo",
						scales: ["x"]
					}
				}
```

and the main opts to:

```js
				cursor: {
					drag: {
						x: true
					},
					sync: {
						key: "moo",
						scales: ["x"]
					}
				}
```

If you keep the ability to drag y on the main chart without it on the ranger you get this odd behavior where moving left and right shows on the ranger even if you are only zooming y so there is some work to do there.

Ps I'm not gonna bother to rebase & squash the pushes until we are settled cause it's a bit of a pita.

EmiPhil:
> If you keep the ability to drag y on the main chart without it on the ranger you get this odd behavior where moving left and right shows on the ranger even if you are only zooming y so there is some work to do there.

I mean "odd" is maybe the wrong word because the ranger is meant to zoom x. The only way it would know not to is for it to see the src chart. But since updateCursor doesn't happen through the sync mechanism it would require another variable for the src drag values. I think that if you are wanting to zoom y on the main chart the recommended opts should probably be to go with the method where it zooms the ranger too.

EmiPhil:
The ranger with everything on reminds be a bit of the mini maps that are in programs like photoshop actually.

I also noticed that single clicking on the sync-cursor demo breaks everything so I'm not too sure what is going on there.

leeoniya:
the locking function is meant to lock the cursor across synced charts. for example to evaluate some legend values across multiple charts (perhaps some scrolled off-screen), or maybe take a screenshot with the cursor positioned at a specific location. the function is click to lock, click again to unlock. these clicks can happen on any of the synced charts with the same key.

i'll need some time to play around and review the changes here. cacheMouse is very perf-sensitive since it the unthrottled mousemove listener and can fire thousands of times per second. any work that can be deferred in it, must be. i'll need to test how much of a perf impact all the extra code has.

EmiPhil:
Would it be possible/reasonable to use the src/sync mechanics in updateCursor?

EmiPhil:
Actually I could put some of the logic into the `mouseMove` function in another raf or set the raf up to call two functions, the caching functionality and the updateCursor functionality

EmiPhil:
See new code where I pulled out the caching logics into their own function. I also only call it in mousemove but now that it is a function we could call it wherever makes sense

EmiPhil:
I think this interface is the cleanest. By adding a second param for src in update cursor we can do the work where we need to without introducing globals. Also I made it so that the uni of the two charts doesn't have to be same - if we are getting our drag command from a src we use the src's uni value and scale our position to that. So you can have a small uni threshold if you want to x/y zoom on the ranger and a much larger one on the main chart. Otherwise they had to be the same so at a uni of, say, 40 on the ranger it was effectively infinity (with this change you could purposefully set the ranger one to infinity if you only wanted x or y).

EmiPhil:
This push makes it so that we only do drag work is the cursor is in bounds on the chart. It can happen that the cursor isn't in bounds if you are syncing by absolute value of x/y but the scales are zoomed different. This also allows the cursor to lock when it is out of bounds, fixing bug where cursor lock wouldn't work when the cursor was out of bounds on one of the sync charts

EmiPhil:
Next up to fix, if you click outside of the zoom range on the ranger it seems to reset the zoom of the main chart. I think this has something to do with the out of bounds logic as well.

src/uPlot.js
```diff
@@ -1762,20 +1811,29 @@ export default function uPlot(opts, data, then) {
...
 
-			if (mouseLeft1 != mouseLeft0 || mouseTop1 != mouseTop0) {
+			// it is possible that the cursor is out of bounds of the chart if it comes via sync,
+			// and it doesn't make sense to do drag work when that is the case (but we still want
+			// to lock the cursor)
+			let mouseXInBounds = mouseLeft0 >= 0 && mouseLeft0 <= plotWidCss;
+			let mouseYInBounds = mouseTop0 >= 0 && mouseTop0 <= plotHgtCss;
+			let mouseInBounds = mouseXInBounds && mouseYInBounds;
+
+			if (mouseLeft1 != mouseLeft0 || mouseTop1 != mouseTop0 && mouseInBounds) {
```
leeoniya, line 1828: think you're missing a set of parentheses here. did you want:

```js
if (mouseInBounds && (mouseLeft1 != mouseLeft0 || mouseTop1 != mouseTop0)) {
```

EmiPhil:
:facepalm: :facepalm: yeppp and that fixes my "next up to fix"

EmiPhil:
Okay, now to break it you have to click within at the veeeeery very top or veeeeery very bottom of the y axis (on the sync-cursor demo).

I think that you have run into this before based on the logic here in cacheMouse:

```js
		if (snap) {
			if (_x <= 1 || _x >= plotWidCss - 1)
				_x = incrRound(_x, plotWidCss);

			if (_y <= 1 || _y >= plotHgtCss - 1)
				_y = incrRound(_y, plotHgtCss);
		}
```

EmiPhil:
The simple fix is to get rid of the check for snap and just force round it in those two cases.

leeoniya:
> I think that you have run into this before based on the logic here in cacheMouse:

this was added because i would frequently not be able to get a mouse event at the very edge (it would be like 0.7px off or something), so it was impossible to select/zoom to the extremes.

EmiPhil:
Hmm gotcha. Getting rid of the snap reintroduces that inability. So I guess fix option 2 is to involve that check in the mouse up area somehow

leeoniya:
> Could we say that the delta of the drag needs to pass a certain threshold?

this might be devicePixelRatio dependent and would likely not be deterministic.

EmiPhil:
Yeah. Even worse actually. I thought I had it with a threshold of 1 but the big chart threshold is closer to 2, so the actual size of the chart will change the threshold

EmiPhil:
By dividing the abs distance by plotWid/HgtCss I think we handle different sized charts/pixel ratios. You have to drag for more than 0.2% of the chart in this case to be dragging.

Instead of just fixing the edge cases (literally! :joy: ) I think this type of threshold is nice because a couple of times I have tried to click the chart but I guess moved the mouse like half a pixel and it zoomed in on the sliver.

EmiPhil:
Latest push adds a check for cursor lock before letting you drag. Because if the chart is locked and you "drag" on one of the other charts it blows up.

EmiPhil:
So I moved the lock check where it makes more sense. This allows you to drag in one chart if it is out of bounds. The next is to change the out of bounds to be less about the size of the chart and more about the min/max of it (if you get them to a point where one chart is zoomed in a different spot than the others and you zoom into an area the others have but isn't currently on screen it won't update them)

EmiPhil:
Actually nevermind on the nevermind above. I get what I meant. The out of bounds thing was actually a golden goose. We just need to check for locks I think...

EmiPhil:
This one works with all of the use cases I can think of right now. It does require extending the pub function to accept extras (I just made that a spread operator for max flexibility in the future). That is maybe not ideal, but if it isSyncReq then we need to know if the src chart is dragging or not. Since that relies on mouse0/1 the only way to do that is via the pub mechanism (or we could have assigned it to self but I don't really like that method) cause those are internal variables

leeoniya:
@EmiPhil i will try to get to this today in more detail.

from a cursory look (no pun intended), can you briefly summarize why the we need additional calls to e.g. `getYPos` if we're already tracking this in move/cache? as i mentioned earlier, if we're live-tracking all this stuff (which i guess is unavoidable if we're gonna have scale-synced stuff).

src/sync.js
```diff
@@ -10,10 +10,10 @@ export function _sync(opts) {
 		unsub(client) {
 			clients = clients.filter(c => c != client);
 		},
-		pub(type, self, x, y, w, h, i) {
+		pub(type, self, x, y, w, h, i, ...extras) {
```
leeoniya, line 13: please don't make this generic. spreads don't de-sugar well and the iife is compiled down to es5 which creates `arguments` bloat and creates extra intermediate arrays, just pass anything extra you need explicitly as i do in the other branch.

image

EmiPhil:
> from a cursory look (no pun intended), can you briefly summarize why the we need additional changes and calls to e.g. getYPos in mouseUp if we're already tracking this in move/cache? as i mentioned earlier, if we're live-tracking all this stuff (which i guess is unavoidable if we're gonna have scale-synced stuff), i feel like there's nothing special we need to do in mouseup.

If you are talking about all of the additions in the `else` part of the check for `if(drag.setScale)` the goal is to figure out what the selection box should be based on what is happening in the other chart. It could be possible to get those values from somewhere else but they are checking for the position of the min/max of the scale of the src chart not the position of the cursor.

The reason that is necessary is to handle the case where you are zooming x/y and one of the charts `drag.setScale = false`. I don't quite remember the behavior before but I think every time you zoomed x on the main chart the ranger would zoom with all of y and vice versa.

EmiPhil:
Behavior without:

uPlotA


Behavior with:

uPlotB

EmiPhil:
So in the before case the ranger is "lying" because it is showing an x/y select that is bigger than what you can really see on the chart. Hopefully the gifs explain better than I could!

leeoniya:
> So in the before case the ranger is "lying" because it is showing an x/y select that is bigger than what you can really see on the chart. Hopefully the gifs explain better than I could!

well, the problem is that it's lying *during* the drag, so you need to do additional checks during mouseUp. it seems like it should be able to account for the lying during dragging, that way at mouseUp everything is already consistent.

EmiPhil:
Hmm good point

EmiPhil:
Oh I think this may be a relatively easy change

EmiPhil:
yeah that is waaaaaaaaay better. Not only did the move get rid of the mouse up reqs but it also simplified the scaling for the drag (if there is a src just set our drag to that drag).

The only think that this doesn't cover is auto scaling on the main chart. So if you zoom into certain areas it will sometimes scale in further but that won't change the setSelect area of the ranger.

Two thoughts though...
1. in most cases if you have a ranger you should maybe be setting auto scale off? Like the whole point of the ranger is to provide the ux for scaling
2. otherwise we would have to publish scale events I think

The use case where 1 would fall apart a bit is for live charts. When it is live you would want auto scaling of new values. But if you were zooming in somewhere with a ranger you would no longer want auto scaling.

Maybe we could provide an option that allows the user to turn off auto scaling after a drag event? (Maybe that should be the default behavior?)

dist/uPlot.d.ts
```diff
@@ -87,6 +87,9 @@ declare class uPlot {
 	/** converts a CSS pixel position (relative to plotting area) to the closest data index */
 	posToIdx(left: number): number;
 
+	/** returns the absolute distance in the scale's units of a given scale */
+	scaleDistance(scaleKey: string): number;
```
leeoniya, line 91: appears to be unused, leftovers?

src/uPlot.js
```diff
@@ -1490,9 +1496,14 @@ export default function uPlot(opts, data, then) {
...
 	}
 
+	function scaleDistance(scale) {
+		return abs(scales[scale].max - scales[scale].min);
+	}
+
 	self.valToIdx = val => closestIdx(val, data[0]);
 	self.posToIdx = closestIdxFromXpos;
-	self.posToVal = (pos, scale) => scaleValueAtPos(scale == xScaleKey ? pos : plotHgtCss - pos, scale);
+	self.scaleDistance = scaleDistance;
```
leeoniya, line 1505: appears to be unused, leftovers?

src/uPlot.js
```diff
@@ -1490,9 +1496,14 @@ export default function uPlot(opts, data, then) {
 		return closestIdx(v, data[0], i0, i1);
 	}
 
+	function scaleDistance(scale) {
```
leeoniya, line 1499: appears to be unused, leftovers?

src/uPlot.js
```diff
@@ -1612,51 +1623,80 @@ export default function uPlot(opts, data, then) {
...
+				let uni = drag.uni;
+	
+				if (uni != null) {
+					let dx = abs(mouseLeft0 - mouseLeft1);
+					let dy = abs(mouseTop0 - mouseTop1);
+	
+					dragX = dx >= uni;
+					dragY = dy >= uni;
+	
+					// force omnidirectionality when both are under uni limit
```
leeoniya, line 1669: the original comment was correct here, i think. we're forcing *uni*directionality here (not omni) by setting *either* dragX or dragY to true when both are false because they're both under the uni threshold.

src/uPlot.js
```diff
@@ -1746,12 +1801,12 @@ export default function uPlot(opts, data, then) {
...
 	function mouseDown(e, src, _x, _y, _w, _h, _i) {
-		if (e == null || filtMouse(e)) {
+		if (src != null || filtMouse(e)) {
 			dragging = true;
 
-			cacheMouse(e, src, _x, _y, _w, _h, _i, true, true);
+			cacheMouse(e, src, _x, _y, _w, _h, _i, true, false);
 
-			if (select.show && (drag.x || drag.y))
+			if (select.show && drag.setScale && (drag.x || drag.y))
```
leeoniya, line 1809: why does selection clearing during mousedown now depend on whether there's a setscale call at mouseup? i think these continue to be orthogonal concerns and should not be conflated. is there a reason for this or is these leftovers from before?

src/uPlot.js
```diff
@@ -1761,21 +1816,38 @@ export default function uPlot(opts, data, then) {
...
+		const isSyncReq = src != null;
+		let shouldDrag;
+
+		if (isSyncReq || filtMouse(e)) {
 			dragging = false;
 
 			cacheMouse(e, src, _x, _y, _w, _h, _i, false, true);
 
-			if (mouseLeft1 != mouseLeft0 || mouseTop1 != mouseTop0) {
+			let locked = cursor.lock && cursor.locked;
```
leeoniya, line 1828: i think it's sufficient to just check cursor.locked and not need to make a variable?

src/uPlot.js
```diff
@@ -1761,21 +1816,38 @@ export default function uPlot(opts, data, then) {
...
+
+			if (_drag == null) {
+				let shouldXDrag = abs(mouseLeft0 - mouseLeft1) / plotWidCss > 0.002;
```
leeoniya, line 1831: this whole new block scares me, especially with hard-coded thresholds. was the original logic of `mouseLeft1 != mouseLeft0 || mouseTop1 != mouseTop0` no longer sufficient to determine if a mousemove happened?

or is this guarding against the case where the synced scales are so different that a mousemove along one chart is insufficient to move 1px on another?

src/uPlot.js
```diff
@@ -1343,8 +1343,8 @@ export default function uPlot(opts, data, then) {
 	const drag = FEAT_CURSOR && cursor.drag;
 
 	let dragX = FEAT_CURSOR && drag.x;
-	let dragY = FEAT_CURSOR && drag.y;;
-
+	let dragY = FEAT_CURSOR && drag.y;
+
```
leeoniya, line 1347: please enable trailing whitespace trimming in your editor.

leeoniya:
still some questions/concerns, but i'm no longer seeing things in disagreeable places, which is great. also could not find anything obviously broken in any of the cursor/zoom/sync demos - cursor locking still works, snapping to primary chart edges still works.

i think it's close!

src/uPlot.js
```diff
@@ -1761,21 +1816,38 @@ export default function uPlot(opts, data, then) {
 		}
 	}
 
-	function mouseUp(e, src, _x, _y, _w, _h, _i) {
-		if ((e == null || filtMouse(e))) {
+	function mouseUp(e, src, _x, _y, _w, _h, _i, _drag) {
```
leeoniya, line 1819: in what situations would _drag during an isSyncReq be different from one asserted from a difference in local e.g. mouseLeft1/mouseLeft0 positions? since mousedown/mousemove/cachemouse is done across all synced charts, wouldnt each chart have local knowledge of whether a drag happend by simply checking the `dragging` boolean?

EmiPhil:
Wrt to "are these leftovers"....uhh probably there is a mix. I think this pr has gone through about 3 or 4 strategies.

I'll walk through it today and try to cut everything that can be, and the things that can't at least provide the edge case they are solving

dist/uPlot.d.ts
EmiPhil: yep

src/uPlot.js
EmiPhil: yep

src/uPlot.js
EmiPhil: yep

src/uPlot.js
EmiPhil: :+1:

src/uPlot.js
EmiPhil: leftover

src/uPlot.js
EmiPhil: :+1:

src/uPlot.js
EmiPhil: This is needed when you are determining drag inside of mouseup instead of at mousemove, no longer needed (though I was thinking select[WIDTH] and select[HEIGHT] are the clearer things to check since that's what ends up being used in all the functions)

src/uPlot.js
EmiPhil: :+1:

src/uPlot.js
EmiPhil: when you are checking if drag is done in mouseup instead of in mousemove; no longer necessary

EmiPhil: actually one of these is necessary to keep cursor locks working, not sure which. Investigating....

src/uPlot.js
EmiPhil: Okay this is the one that causes a cursor lock issue when you remove it. But I think it is because the select isn't being cleared. So this is a bit of a hack to compensate for something else. Investigating more...

src/uPlot.js
EmiPhil: This is not the one, see comment in other review

EmiPhil:
Okay that cleans up pretty much everything.

EmiPhil:
Acutally in the mouse up block there that second if is now kinda useless...

EmiPhil:
Oh this breaks auto scaling actually

EmiPhil:
Okay! I really like this. Instead of using `dragX` or `dragY` which may or may not be the same as `drag.x` and `drag.y` in mouseUp we use `drag.x` and `drag.y` directly. What that means is that if you are doing simple zooms autoscaling works as expected.

But if you are doing x and y dragging then auto scaling is effectively disabled because it will always set the scales for each axis. This allows the zooming to work really intuitively for those more complex scenarios (I think). The zoom-ranger-xy demo really shows it in action if you go back and forth between uni zooming x and y on the main chart. But it still auto scales when you load in data, which is perfect for the first render and in the case of live charts. (I think that as soon as you are zooming x and y it doesn't make sense after a drag to auto zoom the axis, you have put that responsibility in the hands of the user doing the dragging).

src/uPlot.js
```diff
@@ -1746,12 +1792,12 @@ export default function uPlot(opts, data, then) {
...
 	function mouseDown(e, src, _x, _y, _w, _h, _i) {
-		if (e == null || filtMouse(e)) {
+		if (src != null || filtMouse(e)) {
 			dragging = true;
 
-			cacheMouse(e, src, _x, _y, _w, _h, _i, true, true);
+			cacheMouse(e, src, _x, _y, _w, _h, _i, true, false);
 
-			if (select.show && (drag.x || drag.y))
+			if (select.show && (dragX || dragY))
```
leeoniya, line 1800: during mousedown, these will always match `drag.x` and `drag.y`, right?

if that's the case, we can leave this line unchanged and actually pull the dynamic `dragX` and `dragY` variables into the local `else` scope (line 1649), since they will no longer be needed anywhere else.

leeoniya:
> (I think that as soon as you are zooming x and y it doesn't make sense after a drag to auto zoom the axis, you have put that responsibility in the hands of the user doing the dragging).

yeah, for sure.

EmiPhil:
Yep and moved

leeoniya:
actually, relying on only `drag.x` and `drag.y` breaks autoscaling in the dygraphs-style zooming :(, (see zoom variations). also single-clicking in that demos does weird things.

EmiPhil:
On branch 618 or b56? 6a8 is garbage I pushed it too eagerly

leeoniya:
i just pulled b56 and zoom variations has the mentioned issues. i don't see how we can really get away from relying on the dynamic/shared dragX and dragY because the autoscale behavior will depend on what actually happened rather than the drag.x and drag.y opts - those should just serve to manage the behavior during dragging. we can just reset dragX and dragY to the original opts in mouseup or something.

EmiPhil:
Okay I got the clicking thing fixed, but I need some clarification for the expected behavior of the dygraphs one. We want it to act completely like the just x and y ones and so auto scale the other axis if necessary after zoom?

leeoniya:
> We want it to act completely like the just x and y ones and so auto scale the other axis after zoom?

well, there's not really a way to auto-scale x, but for y, yes.

EmiPhil:
Okay yeah. And I think the limitation on auto scaling has more to do with `sync` than with the x / y.

EmiPhil:
This all seems to work well enough? Behavior for non sync charts behaves without limitations, but if we are syncing then we just use the drag.x and drag.y values to prevent some of the sync charts from auto scaling while others don't.

EmiPhil:
Most recent push brings back the setSelect function that should have never been nuked

leeoniya:
> This all seems to work well enough?

there's definitely some oddities that happen with syncing exactly to match the src selection during unidirectional dragging. in the sync-cursor demo, if you drag along x in the smaller chart that is auto-scaled to have a min of 10%, you'll see that 10% min reflected in the main chart and will frequently end up with an empty main chart after zooming - but you only dragged x!

perhaps we do need to pub dragX and dragY to all subscribers so they know if they need to match an explicit x and/or y drag.

also, `timezones-dst` demo seems to have lost the sync'd setSelect?

EmiPhil:
For the timezone-dst I think that should be fixed by d6c7304. I accidentally deleted the setSelect hook cause I thought it was leftovers.

For the second one, it is odd but myabe not wrong? Like if you think about it they are supposed to be synced together but at render they are out of sync. So when you zoom the x of the zoomed in chart you are effectively zooming x/y on the non synced charts.

EmiPhil:
I guess my thought is that if you are explicitly syncing the x/y key of two charts, you would expect the two scales to reflect each other. I think in this sync cursor demo that isn't the intent - it should really only sync the x.

EmiPhil:
And yeah taking out the "y" key and the uni/y drag in the cursor opts in that demo behaves more how you would want this type of chart to behave. (What does it even mean to sync y across different units? This is why we left the ability to have y be relative imo)

EmiPhil:
12d0ddd includes an updated cursor-sync demo to reflect the changes I just mentioned

leeoniya:
yes you're right, i didn't actually mean for my original branch to permanently include those sync-cursor changes [1] in the demo, it was more just playing around. i'll give this a final review later today.

[1] https://github.com/leeoniya/uPlot/commit/e3a877e392cdcbd6335a25ef934fac62eff3bd39#diff-6fc4e995c333753aaa6e5d2c73a8660e

leeoniya:
since `scales: ["x"],` is the default behavior now, you can probably skip any changes to sync-cursor.

EmiPhil:
:+1: 

And the change in 6d0ee4a is to fix a visual issue where if you are only syncing one or the other axis it wasn't always showing the whole select on the other chart.

leeoniya:
5f265d1 was a winner :)

outstanding work! i will give you 20% of profits from uPlot. unfortunately, 20% of $0 is still $0 🤣 

i'm gonna do some followups like merging the rangers into one demo and other minor stuff.

Issue #219, opened 2020-05-14
Auto X / Y Zoom


EmiPhil:
Reference drawing:

![image](https://user-images.githubusercontent.com/11808725/81958724-9a6dac80-95cb-11ea-9dee-118fe658265d.png)

Say you are trying to zoom from the point Mo.

What I'd like to do:

1. If the cursor is in the invisible green band of height x pixels, zoom in on the X axis.
2. If the cursor is in the invisible purple band of width x pixels, zoom in on the Y axis.
3. If the cursor is in the invisible burgundy intersection of height/width x pixels, don't zoom at all
4. If the cursor is in any of the four areas outside of the bands, zoom in on both the X and Y axis.

Hopefully that explanation makes sense!

Is this something that should be done via a plugin or would you consider it being part of the lib?

Thanks.

EmiPhil:
A while ago I threw together a quick proof of concept while evaluating AM charts, so this kind of shows what I mean interactively (but doesn't work super great).

https://codepen.io/Akaritzu/pen/RwPGREx

leeoniya:
dygraphs has something like this too (try on the home page): http://dygraphs.com/

except they only flip between either x or y but never both. from the amcharts codepen, i think having all 3 variants causes too much UI flashing during the selection...and you dont really save that much time over just doing an x zoom followed by a y zoom.

i might be open to having the dy-graphs style if it doesnt add too much code or option complexity. i don't think we need to go as far as adding thresholds but just use whichever is greater of dx or dy to toggle between x or y scale. that means another option, e.g. `cursor.drag: {x: true, y: true, bidi: false}`

that being said, if what you need is the 3-variant flavor, having the 2-variant in the core won't help you much - you'll probably need to make a plugin that implements all the selection logic.

EmiPhil:
So doing it both ways is fairly straightforward if it works for you.

Basically instead of setting omni to a boolean I used a number:

`omni = 0` means use the default behavior
`omni = 1` means use omni with no threshold (the dygraph way)
`omni > 1` means use omni with a threshold of omni (the way I'm looking for)

leeoniya:
not bad, actually.

except that the PR breaks unidirectional zooming 🤣

EmiPhil:
:facepalm:

Maybe the next thing to look at is #72

leeoniya:
that's a one large can of worms, my friend. even #175 wouldn't have saved you in this case.

EmiPhil:
[...]

PR #221, opened 2020-05-14
Omni x/y zoom

EmiPhil:
#219


leeoniya:
huh, i guess i merged e4ee6f5, but there was a later one. so this did not get marked as merged.

i think all the changes are merged.

PR #242, opened 2020-05-25
Tiny type definition fix for series.point options

mrkurt:
The fields in `point` are all optional, this just fixes the type definition to match.


leeoniya:
awesome, thanks!

PR #247, opened 2020-05-27
allow setting a min dist for drag. close #245

### LINKED ISSUE #245: Cancel Drag / Min Drag Distance
It would be nice to be able to specify an option to only start dragging if the mouse has moved a certain distance (and if the mouse goes past that distance and then back in, cancel the drag instead of zooming in on the little area)

EmiPhil:


dist/uPlot.d.ts
```diff
@@ -237,6 +237,8 @@ declare namespace uPlot {
 			x?: boolean; // true
 			/** toggles dragging in along y */
 			y?: boolean; // false
+			/** sets min distance that drag must pass to execute */
```
leeoniya, line 240: `min drag distance threshold` is shorter and i think clear enough

leeoniya:
haven't reviewed this too thoroughly yet. let's rebase this on top of #248 once that lands.

src/uPlot.js
```diff
@@ -1658,29 +1658,38 @@ export default function uPlot(opts, data, then) {
...
+	
+						// force unidirectionality when both are under uni limit
+						if (!dragX && !dragY) {
+							if (dy > dx)
+								dragY = true;
+							else
+								dragX = true;
+						}
 					}
+				} else if (drag.x && drag.y && dragX != dragY) {
```
leeoniya, line 1683: you dont need to change this here (i can do it as a follow-up) but for future reference, i don't like cuddled `else` blocks.

src/uPlot.js
```diff
@@ -1658,29 +1658,38 @@ export default function uPlot(opts, data, then) {
...
 			else {
 				// setSelect should not be triggered on move events
-				let uni = drag.uni;
+				let dx = abs(mouseLeft0 - mouseLeft1);
+				let dy = abs(mouseTop0 - mouseTop1);
 
-				if (uni != null) {
-					let dx = abs(mouseLeft0 - mouseLeft1);
-					let dy = abs(mouseTop0 - mouseTop1);
+				let dragX = drag.x && dx >= drag.dist;
```
leeoniya, line 1664: so, we already have global dragX and dragY vars which these are masking - the name clash is confusing. would it not be possible to reuse the existing globals here? their purpose seems identical.

demos/zoom-variations.html
```diff
@@ -27,9 +27,14 @@
...
 				['X only', { x: true, y: false }],
 				['Y only', { x: false, y: true }],
+				['X only + min distance', { x: true, y: false, dist: 50 }],
+				['Y only + min distance', { x: false, y: true, dist: 50 }],
 				['X or Y (adaptive)', { x: true, y: true, uni: Infinity }],
+				['X or Y (adaptive) + min distance', { x: true, y: true, uni: Infinity, dist: 50 }],
 				['X or Y (omni)', { x: true, y: true }],
+				['X or Y (omni) + min distance', { x: true, y: true, dist: 50 }],
 				['X or Y (adaptive + omni)', { x: true, y: true, uni: 50 }],
+				['X or Y (adaptive + omni) + min distance', { x: true, y: true, uni: 50, dist: 50 }],
```
leeoniya, line 37: i would place all the `dist` demos at the end (maybe even after some `<h2>` for clarity, otherwise it might feel like a bug that the selection doesnt start immediately.

also, while dist: 50 is obvious, it's pretty crazy. i think the majority case would be to prevent accidental drags, so something like 10px is much more reasonable without preventing zooms into pretty large areas.

src/uPlot.js
```diff
@@ -1658,29 +1658,38 @@ export default function uPlot(opts, data, then) {
...
+							if (dy > dx)
+								dragY = true;
+							else
+								dragX = true;
+						}
 					}
+				} else if (drag.x && drag.y && dragX != dragY) {
+					// if omni with no uni then both dragX / dragY should be true if either is true
+					dragX = true;
+					dragY = true;
```
leeoniya, line 1686: this can be `dragX = dragY = true;`

src/uPlot.js
leeoniya: i know this is a bit less efficient, but to me `&& (dragX || dragY)` would be a bit more clear.

leeoniya: @EmiPhil i see `dragX != dragY` is still in place. was there an issue with `(dragX || dragY)`, or is this still WIP?

leeoniya:
thanks!

PR #248, opened 2020-05-27
close #244

### LINKED ISSUE #244: mouse zooming at the edges
in the candelstick demo, if you use mouse to zoom, sometimes the edge ticks would be excluded or missed. or if you hold the mouse left button and swing the mouse out of canvas, the selection range could be random. ideally, when mouse cursor is out of range, every ticks along that direction should be included?

EmiPhil:
unless we want proxthresh to be a setting #244

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -1868,11 +1868,57 @@ export default function uPlot(opts, data, then) {
 	}
 
 	function mouseLeave(e, src, _x, _y, _w, _h, _i) {
-		if (!cursor.locked && !dragging) {
+		if (!cursor.locked) {
+			let _dragging = dragging;
+
+			if (dragging) {
+				// handle case when mousemove aren't fired all the way to edges by browser
+				let dLft = mouseLeft1;
+				let dRgt = plotWidCss - mouseLeft1;
+				let dTop = mouseTop1;
+				let dBtm = plotHgtCss - mouseTop1;
+
+				let snapX = true;
+				let snapY = true;
+				let snapProx = 10;
+
+				if (dragX && dragY) {
+					// maybe omni corner snap
+					snapX = mouseLeft1 <= snapProx || mouseLeft1 >= plotWidCss - snapProx;
+					snapY = mouseTop1  <= snapProx || mouseTop1  >= plotHgtCss - snapProx;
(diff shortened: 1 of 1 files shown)


src/uPlot.js
```diff
@@ -1868,11 +1868,55 @@ export default function uPlot(opts, data, then) {
...
+			let _dragging = dragging;
+
+			if (dragging) {
+				// handle case when mousemove aren't fired all the way to edges by browser
+				let dLft = mouseLeft1;
+				let dRgt = plotWidCss - mouseLeft1;
+				let dTop = mouseTop1;
+				let dBtm = plotHgtCss - mouseTop1;
+
+				let xProx = true;
```
leeoniya, line 1881: if these are booleans, let's use `snap` instead of `prox`. e.g. `snapX`.

src/uPlot.js
```diff
@@ -1868,11 +1868,55 @@ export default function uPlot(opts, data, then) {
...
+				let yProx = true;
+				let proxThresh = 10;
```
leeoniya, line 1883: `snapProx`

src/uPlot.js
```diff
@@ -1868,11 +1868,55 @@ export default function uPlot(opts, data, then) {
...
+
+				if (dragX && dragY) {
+					// maybe omni corner snap
+					xProx = mouseLeft1 <= proxThresh || mouseLeft1 >= plotWidCss - proxThresh;
+					yProx = mouseTop1  <= proxThresh || mouseTop1  >= plotHgtCss - proxThresh;
+				}
+				
+				if (dragX && xProx) {	
+					let xMin = min(dLft, dRgt);
+					if (xMin == dLft)
```
leeoniya, line 1893: please make sure any leading `if ()` blocks have an empty line before them. more legible to me than having `let` immediately adjacent, especially with identical alignment.

leeoniya:
> unless we want proxthresh to be a setting

let's wait till at least 1 person complains ;)

looks good. please address the tweaks, exclude the dist builds and squash into a better commit message:

```
snap selection to plot edges on mouseleave. close #244
```

EmiPhil:
should be gtg

Issue #250, opened 2020-05-29
Canvas onclick and ondblclick broken


danekasper:
Init hook bindings of canvas with onclick and ondblclick do not fire click clicking or double clicking on the canvas. 

eg. https://leeoniya.github.io/uPlot/demos/zoom-fetch.html and 

```js
hooks: {
  init: [
    u => {
      u.ctx.canvas.ondblclick = e => {
        console.log("Fetching data for full range");
        u.setData(data);
      }
    }
  ]
} 
```

I see work is currently ongoing in this space, so it may be known issue (https://github.com/leeoniya/uPlot/commit/8c3e6620287e18ff2d022fbd8841d5d7fd455c58)

Edit: Clicking does work on the axis/title area, leading me to believe perhaps there was a change to how the plot area is rendered that is capturing the click event.

leeoniya:
yeah, there's a bit of an ongoing chat about dblclick and how it's expected to behave during `sync`. i think it's pretty much settled as of a few minutes ago.

https://github.com/leeoniya/uPlot/commit/d5eed15ae43fef3bb9dae914fd6ea89353b1e895#r39495032

danekasper:
event bindings should attach to u.root.querySelector(".over") not u.ctx.canvas
Thanks!

PR #251, opened 2020-05-29
Update zoom-fetch demo to accept ondblclick

danekasper:
canvas.ondblclick event doesn't fire when clicking on the plot, updated to use the div.over element accessible by u.ctx.canvas.nextElementSibling #250

### DIFF
--- a/demos/zoom-fetch.html
+++ b/demos/zoom-fetch.html
@@ -34,7 +34,7 @@
 				hooks: {
 					init: [
 						u => {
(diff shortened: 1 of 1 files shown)


leeoniya:
pretty much all event bindings should not move to the `u.root.querySelector(".over")` element. see if that works for you.

leeoniya:
101d488 updates the zoom-fetch demo to do this

danekasper:
Updated on master

PR #256, opened 2020-06-09
check for legend/cursor point before splicing. close #255

### LINKED ISSUE #255: delSeries throws if legend.show = false
`TypeError: legendRows.splice(...)[0] is undefined`

``` js
  1127 | 
  1128 | function delSeries(i) {
  1129 | 	series.splice(i, 1);
> 1130 | 	 legendRows.splice(i, 1)[0][0].parentNode.remove();
       | ^  1131 | 	 cursorPts.splice(i, 1)[0].remove();
  1132 | 
  1133 | 	// TODO: de-init no-longer-needed scales?
```

I think this is a simple fix to also check if legendRows exists

EmiPhil:


src/uPlot.js
```diff
@@ -381,8 +381,8 @@ export default function uPlot(opts, data, then) {
 
 	function delSeries(i) {
 		series.splice(i, 1);
-		FEAT_LEGEND && legendRows.splice(i, 1)[0][0].parentNode.remove();
-		FEAT_CURSOR && cursorPts.splice(i, 1)[0].remove();
+		FEAT_LEGEND && legendRows[i] && legendRows.splice(i, 1)[0][0].parentNode.remove();
```
EmiPhil, line 384: This could also be 
```js
		FEAT_LEGEND && showLegend && legendRows.splice(i, 1)[0][0].parentNode.remove();
```

but I'm not sure cursorPts has a similar bool to check so I went with this so that they are consistent.

leeoniya: yeah, i was gonna ask this.

> `FEAT_LEGEND && showLegend`

i like this more, since we already have the dedicated flag and it's used evrywhere.

> but I'm not sure cursorPts has a similar bool to check so I went with this so that they are consistent.

updateCursor uses `FEAT_CURSOR && cursorPts.length > 1`, so let's do the same here.

EmiPhil: :+1:

leeoniya:
thanks!

PR #260, opened 2020-06-12
Add opts.classMap to customize class names generated by uPlot

mneumann:
classMap is a function taking a class name and returning a class name. Apply
this function to every class name generated by uPlot.

This fixes one issue when using uPlot with Bulma.css. Bulma has a global .select
class that interferes with uPlot's .select class. The effect is an annoying blue
"drop-down" symbol drawn on the plot area.

The default for opts.classMap is the identity map.


leeoniya:
hey @mneumann thanks for the PR.

i'm not sure this is worth having per-instance.

instead, how about exposing a static global setter like `uPlot.classMap(c => c)` which can override some internal `let _classMap = c => c` middleware. and just modify all the existing dom functions to pass the classes through it.

this should compress better and avoid creating additional decorator functions per plot instance.

another API might be to make a more general `uPlot.config({classMap: u => u})`. this could be expanded in the future to adjust other global defaults without growing the list of individual config setters.

leeoniya:
after some testing and thinking, i decided to simply prefix uPlot's classnames to reduce the probability of collisions. some classnames have also changed to be more clear (this will be a breaking change for v1.1 in non-default setups, but the changes are simple and purely mechanical). additionally, i've pulled all classnames into a separate file to make it easily changeable in custom builds: https://github.com/leeoniya/uPlot/blob/master/src/domClasses.js

mneumann:
@leeoniya Thanks for taking care of this! Having a "u-" prefix works perfect for my use-case!

PR #262, opened 2020-06-22
Fix TypeScript definitions for plugin hooks

jdesgats:
For plugins, it is possible to pass either a bare function or an array
of functions. The TypeScript API definition was enforcing arrays only.

The `Hooks[P][0]` part for removing the array feels a bit hacky, but it keep the change smaller. Another way to define this would be to make a private interface without the array part, and define both `Hooks` and `PluginHooks` based on that. Happy to do that instead if you think it is clearer.

### DIFF
--- a/dist/uPlot.d.ts
+++ b/dist/uPlot.d.ts
@@ -189,7 +189,7 @@ declare namespace uPlot {
 		plugins?: {
 			/** can mutate provided opts as necessary */
 			opts?: (self: uPlot, opts: Options) => void;
-			hooks: Hooks;
+			hooks: PluginHooks;
 		}[];
 	}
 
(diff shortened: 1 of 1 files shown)


leeoniya:
@jdesgats thanks for the fixup!

> Another way to define this would be to make a private interface without the array part, and define both Hooks and PluginHooks based on that.

can you paste here what that would look like?

jdesgats:
This is what it would look like:

```ts
interface HooksDescription {
	/** fires after opts are defaulted & merged but data has not been set and scales have not been ranged */
	init?:       (self: uPlot, opts: Options, data: AlignedData) => void;

	/** fires after any scale has changed */
	setScale?:   (self: uPlot, scaleKey: string) => void;

	/** fires after the cursor is moved (debounced by rAF) */
	setCursor?:  (self: uPlot) => void;

	/** fires after a selection is completed */
	setSelect?:  (self: uPlot) => void;

	/** fires after a series is toggled or focused */
	setSeries?:  (self: uPlot, seriesIdx: number, opts: Series) => void;

	/** fires after data is updated updated */
	setData?:    (self: uPlot) => void;

	/** fires after the chart is resized */
	setSize?:    (self: uPlot) => void;

	/** fires at start of every redraw */
	drawClear?:  (self: uPlot) => void;

	/** fires after all axes are drawn */
	drawAxes?:   (self: uPlot) => void;

	/** fires after each series is drawn */
	drawSeries?: (self: uPlot, seriesKey: string) => void;

	/** fires after everything is drawn */
	draw?:       (self: uPlot) => void;

	/** fires after the chart is fully initialized a.nd in the DOM */
	ready?:      (self: uPlot) => void;

	/** fires after the chart is destroyed */
	destroy?:    (self: uPlot) => void;
}

export type Hooks = { [P in keyof HooksDescription]: HooksDescription[P][] }
export type PluginHooks = { [P in keyof HooksDescription]: HooksDescription[P] | HooksDescription[P][] }
```

`Hook` members must be arrays while `PluginHooks` can be either function or array.

leeoniya:
let's go with the pasted variant for clarity. thanks!

jdesgats:
Sure, I just updated the commit.

leeoniya:
thanks!

PR #277, opened 2020-07-21
Upgrade benchmark to Chart.js 3.0.0-alpha.2

benmccann:
This is about 25% faster on my machine

### DIFF
--- a/bench/Chart.js-next.html
+++ b/bench/Chart.js-next.html
@@ -17,10 +17,8 @@
 	<body>
 		<h2 id="wait">Loading lib....</h2>
 
-		<script src="https://cdn.jsdelivr.net/npm/luxon@1.22.0"></script>
-		<script src="https://cdn.jsdelivr.net/npm/chart.js@3.0.0-alpha"></script>
-		<script src="https://cdn.jsdelivr.net/npm/chartjs-adapter-luxon@0.2.0"></script>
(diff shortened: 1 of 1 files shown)


leeoniya:
thanks, about 13% faster here, ~315ms -> ~275ms. i'll merge this when i have some more time to do a full bench run.

leeoniya:
@benmccann is there a way to improve the cursor performance, at least outside the plotting area where the legend toggles are?

when i do these benchmarks i notice a ton of time is being spent on me simply moving the cursor to the legend to toggle it. even a short distance whch comes in from the edge and does not cross the plotting area is crazy expensive.

on the one hand i want to [tediously] exclude this to have apples to apples. on the other hand this slowdown will affect all practical use cases IRL, so i dont feel like you guys deserve the pass here.

benmccann:
Thanks for the suggestion! I've added that to the suggested list of improvements for v3. I know our tooltip performance could be better, but I've never actually seen a user-visible impact from it, so have never prioritized it. Plus I'm trying to spend less time of the charting stuff, so doubt that I'll get around to implementing it, but perhaps someone else will pick it up.

leeoniya:
okay, thanks.

i guess i'll just try to come in from the edge and also include any intra-legend mousemoves as part of the toggle 6x metric. even this would be quite generous since most toggle cases in real life will have multi-factor more mousemoves before reaching the legend.

PR #278, opened 2020-07-22
Adds check for legendRows.length

nicwestvold:
If `legend.show` option is set to false, then `legendRows` is an empty array
and the call to `legendRows[i][0]` on line #1499 will cause an error,
since `legendRows[i]` is undefined.

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -1495,7 +1495,7 @@ export default function uPlot(opts, data, then) {
 	function _alpha(i, value) {
 		series[i].alpha = value;
 
-		if (FEAT_LEGEND && legendRows)
+		if (FEAT_LEGEND && showLegend)
 			legendRows[i][0].parentNode.style.opacity = value;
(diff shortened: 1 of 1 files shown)


leeoniya:
hey @nicwestvold thanks for the PR!

might be better to use `if (FEAT_LEGEND && showLegend)` as is done in every other place? this one instance might be a leftover from before `showLegend` existed as a cached internal var.

leeoniya:
also, please squash (sorry!)

nicwestvold:
@leeoniya, thanks for the quick response and pointing me in the right direction! `showLegend` looks like it's exactly what I wanted.

nicwestvold:
> also, please squash (sorry!)

Don't be sorry! I wasn't sure how to contribute and was doing it wrong. Thank you for walking me through it!

leeoniya:
no problem. thanks for the fix!

Issue #269, opened 2020-07-09
cursor.snap


leeoniya:
so that the cursor only moves to exact x indices.

would be useful to add this mode, since visually it's more obvious than just noticing the cursor points snapping.

the whole updateCursor should be conditioned on the index changing, similar to legend updates:

https://github.com/leeoniya/uPlot/blob/ee065f6022c6380dbc5e8389260ea334e47c318e/src/uPlot.js#L1622-L1623

cc @EmiPhil

leeoniya:
i did a quick prototype of `cursor.snap: true` in https://github.com/leeoniya/uPlot/tree/cursor.snap.

some general thoughts:

it's a bit odd to have a crosshair cursor snap in the x direction (to the datapoints), but act smoothly in the y direction, but i guess there's not much to be done about that. maybe if `snap: true`, the `setCursor` hook should not fire unless the snap location changes? dunno.

i don't particularly like the purpose-specific nature of a `snap: true` API, since it should be possible to snap generally, and not specifically to data but to arbitrary points on a scale, if these points happen to come from data then so be it.

i think a general `cursor.snap` API would look something like below, where we snap to the returned values on along the indicated scale.

```js
cursor: {
    snap: {
        x: {
            scale: 'x',
            values: self => self.data[0],
        },
        y: {
            scale: 'myScale',
            values: self => [2,4,6,8,10],
        },
    }
}
```

ScottKevill:
> it's a bit odd to have a crosshair cursor snap in the x direction (to the datapoints), but act smoothly in the y direction, but i guess there's not much to be done about that. maybe if `snap: true`, the `setCursor` hook should not fire unless the snap location changes? dunno.

I had a play with this, and agree it doesn't feel natural with the smooth y-direction. However, snapping the X-direction and then snapping the Y-direction to the closest series data point from the X snap does feel very intuitive, at least on the Area Fill demo. So much so that it feels like it could be the default behaviour.

Undecided when it comes to gaps though, eg. the Nearest Non-Null demo. Each time I try it, I change my mind. It feels acceptable as above, but could probably only be improved by a 2D min-distance search.

> i don't particularly like the purpose-specific nature of a `snap: true` API, since it should be possible to snap generally, and not specifically to data but to arbitrary points on a scale, if these points happen to come from data then so be it. a general `cursor.snap` API would look something like below, where we snap to the returned values along the indicated scale.

[...]

PR #291, opened 2020-08-20
Fix drag-select + cursor.move + cursor.drag.dist

ScottKevill:
A solution for the problems mentioned in https://github.com/leeoniya/uPlot/issues/269#issuecomment-676821442


leeoniya:
👍

PR #321, opened 2020-10-01
Upgrade Chart.js to beta.5

benmccann:


leeoniya:
any expected perf or size wins?

benmccann:
Not that I'm aware of, but I've kind of fallen off in terms of helping with development lately and haven't kept up

Issue #279, opened 2020-07-27
Rotate chart


choma:
We are testing uPlot in a project that draws hundreds of thousands of data points, and it works great. A requirement of the project is that the chart needs to be "rotated", I mean, the x axis is the vertical axis, like this [Echart example](https://echarts.apache.org/examples/en/editor.html?c=line-y-category).

I'm currently trying to do this by rotating labels and then rotating the canvas. It kind of works, but [label rotation is only supported in the x axis](https://github.com/leeoniya/uPlot/blob/a80c61962ca4f462a28880122bd94bf58831f7cc/src/uPlot.js#L1215).

Is there an easier way to do this?

leeoniya:
hey @choma 

that's interesting. can you post a jsfiddle example with your config and some basic data. initially i was gonna dismiss this whole thing as impossible, but thinking about it more makes me wonder how far it can get without issues by simply overriding axis.side (i assume?).

if the only issue is the side restriction for label rotation, this should be fairly easy to address.

leeoniya:
what i was wondering if this "rotation" can be supported natively by simply swapping some `scale` and `side` options.

it doesn't quite: https://jsfiddle.net/qyjsbLzm/1/

i'm not sure how many internal assumptions would need to be tweaked to make this work, but it seems like a good thing to look into. i would prefer to solve rotation properly (like this) instead of adding label rotation to other axes as a half-hack that still leaves the interaction completely busted.

choma:
I was trying something like this:

https://jsbin.com/zihuxaqumo/edit?js,output

The drawing looks ok, but the interactions is obviously broken

leeoniya:
since this is a fairly niche use-case and the changes needed are likely not trivial, chart rotation is not at the top of my priority list; i'm going to release v1.1 hopefully in the next few days, then start plugin cleanup & #184 next.

you may have to settle for a horizontal x until i get to this, or use another lib if chart rotation is critical to your app.

choma:
Thanks for the quick answer @leeoniya.
We will evaluate our options, but we are going to keep an eye on this library in any case.

1nv1:
The possibility of rotate the chart would be great! That feature is very used in many academics and hobbyists areas. Regards.

leeoniya:
the thing in my mind that this would be most useful for is actually horizontal bars. where the vertical "x" serves as a category axis and is often used in bar race charts. eg like a rotated version of the already-crowded perf chart [1].

@1nv1 

[...]

PR #336, opened 2020-10-19
Adds rotated functionality

seraco:
The purpose of this pull request is to enable plotting the main scale in the vertical axis and the rest of scales in the horizontal axis.


leeoniya:
hi @seraco ,

thanks for the PR, this is quite the chunk of work! 👍 

a few initial comments before a larger review:

1. this may bitrot a little, since i'd like to fix a couple outstanding issues and cut a release pretty soon, and this is a large change which will need a lot of testing.
2. please leave the `/dist` builds out of the PR to make reviewing easier. i will build & commit them myself.
3. this appears to take the scortched-earth/touches-everything approach i was hoping to avoid by leveraging the browser's native `transform: rotateZ()` and simply poking the parts that relate to bounding box and mouse interaction math, plus label rotation compensation. have you seen and/or considered this instead: https://github.com/leeoniya/uPlot/issues/279#issuecomment-707381731?
4. if this approach is chosen, i'd prefer to keep most of the internal X/Y nomenclature consistent with dependent/independent rather than vertical/horizontal association. which would mean renaming things like dragX/dragY to dragV/dragH or something, but this can be done in a follow-up to make reviewing easier. alternatively, if the rotateZ() approach works out better, nothing needs to be renamed internally, which is a huge plus and reduces cognitive load (imo).

seraco:
Hey @leeoniya,

Thank you for your feedback.

The reason why I finally went with this approach is because my previous attempts at using canvas rotation or CSS transform failed.

Unnecessary files from `/dist` folder have been removed from this PR now.

Finally, I hope this contribution can be useful to the project and can help solving some of the issues described in #279.

leeoniya:
> The reason why I finally went with this approach is because my previous attempts at using canvas rotation or CSS transform failed.

would you mind expanding on what the main issues were?

seraco:
> > The reason why I finally went with this approach is because my previous attempts at using canvas rotation or CSS transform failed.
> 
> would you mind expanding on what the main issues were?
> 
> i imagine with a CSS transform the DOM layout will get screwy and would need to be compensated for with more CSS.

Exactly, with CSS transform DOM layout was broken.

However, the main issues were related to user interaction. All the logic for handling zooming and updating cursor values for legend was no longer working.

leeoniya:
@seraco i'm going to go a slightly different route with this. the idea is to add scale direction and scale orientation options, which together will provide what's needed. i've landed the first part of this (scale direction) in https://github.com/leeoniya/uPlot/commit/6ab82eb4b490e83d677a8f671a5ca786ef3464ed. the rest will land by the end of the year. i'll peek back at this PR if i need some inspiration. thanks!

PR #347, opened 2020-10-28
axis.textAlign

### LINKED ISSUE #345: Alignment of axis values for inset/overlay axes
Hey @leeoniya, after reading [your issue for strategies for auto axis.size](https://github.com/leeoniya/uPlot/issues/333) I thought about one possible strategy which I _think_ would be cheap to implement.

## Current visual
The problem: when the axis size is not enough for the axis value length, the value gets clipped:

ss1

## Desired
I managed to achieve the following visual using a negative `gap` along with `size: 0`:

ss2

```js
axes: [
    {
        size: 0,
        gap: -32
    },
    {
        size: 0,
        gap: -50
    }
]
```

Even tough the values are still getting clipped, we could fix it **by rendering the text aligned to the left instead of right**:

Notice how the `CPU Usage` axis values are aligned to the left, which makes longer values like `100%` still fully visible.

Also, and when using `side: 1 // right side`:

ss3

Then the text should be aligned to the right instead of left.

## Suggested solution

Maybe introduce a property named `align` for the `Axis` type, which would allow us to achieve this desired visual. Since this should not require doing any additional renders, it would serve as one decent solution to the axis size problem.

joaopaulobdac:
This is my shot at adding the textAlign prop for axes. Here's the visual with negative gap values and `textAlign: 1 // left` on the y axis (tested on the `1.2.2` tag branch):

ss

Closes #345

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -1295,7 +1295,9 @@ export default function uPlot(opts, data, then) {
 
 			ctx.font         = axis.font[0];
 			ctx.fillStyle    = axis.stroke || hexBlack;									// rgba?
-			ctx.textAlign    = angle > 0 ? LEFT :
+			ctx.textAlign    = axis.align == 1 ? LEFT :
+			                   axis.align == 2 ? RIGHT :
+			                   angle > 0 ? LEFT :
 			                   angle < 0 ? RIGHT :
 			                   ori == 0 ? "center" : side == 3 ? RIGHT : LEFT;
 			ctx.textBaseline = angle ||
--- a/dist/uPlot.d.ts
+++ b/dist/uPlot.d.ts
@@ -478,6 +478,9 @@ declare namespace uPlot {
 		/** values rotation in degrees off horizontal (only bottom axes w/ side: 2) */
 		rotate?: number | ((self: uPlot, values: Array<string|number>, axisIdx: number, foundSpace: number) => number);
 
+		/** text alignment of axis values - 1: left, 2: right */
+		align?: 1 | 2;
+
 		/** gridlines to draw from this axis' splits */
 		grid?: {
 			/** grid on/off */


leeoniya:
thanks!

i'd prefer `align` instead of `textAlign`.

otherwise, LGTM.

leeoniya:
also please squash into a meaningful commit msg. something like:

```
add axis.align for values
```

joaopaulobdac:
> i'd prefer `align` instead of `textAlign`.

At first I went with that too, but are you sure it wouldn't be confused with aligning the axis itself instead of the axis value? Your call :+1:

leeoniya:
yeah, let's go with `align`.

one can argue that axis alignment is already `axis.side`, plus the ambiguity ship has sailed with `rotate` affecting values and not the axis itself :D, and many other places in the lib. either way textAlign is not unambiguous anyways, maybe it's for `axis.label` alignment? 🤷 naming things is hard, RTFM 🤣

src/uPlot.js
```diff
@@ -1295,11 +1295,14 @@ export default function uPlot(opts, data, then) {
 
 			ctx.font         = axis.font[0];
 			ctx.fillStyle    = axis.stroke || hexBlack;									// rgba?
-			ctx.textAlign    = angle > 0 ? LEFT :
+			ctx.textAlign    = axis.align == 1 ? LEFT :
+					           axis.align == 2 ? RIGHT :
+					           angle > 0 ? LEFT :
 			                   angle < 0 ? RIGHT :
-			                   ori == 0 ? "center" : side == 3 ? RIGHT : LEFT;
+							   ori == 0 ? "center" :
```
leeoniya, line 1302: please leave the unrelated re-formatting here and below as it was. it was aligned it on purpose (pardon the pun?).

joaopaulobdac:
HAHA true, if the entire lib were super verbose using names like `axisValueTextAlign` it would probably be 30KB heavier too :rofl: 

btw, how should I squash these commits into one without breaking everything? I've never done it actually

leeoniya:
i typically use git via TortoiseGIT, and the squashing there is as easy as selecting the desired commits in the log and choosing "combine".

via cli it look as simple as every other git command 🤣 :

https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git/5201642#5201642

after squashing you need to do a force-push.

leeoniya:
you can also just open up another PR with just the final changes and a single commit msg.

src/uPlot.js
```diff
@@ -1295,11 +1295,13 @@ export default function uPlot(opts, data, then) {
...
-			                   angle < 0 ? RIGHT :
-			                   ori == 0 ? "center" : side == 3 ? RIGHT : LEFT;
+			ctx.textAlign    = axis.align == 1 ? LEFT :
+							   axis.align == 2 ? RIGHT :
+							   angle > 0 ? LEFT :
+							   angle < 0 ? RIGHT :
+							   ori == 0 ? "center" : side == 3 ? RIGHT : LEFT;
 			ctx.textBaseline = angle ||
-			                   ori == 1 ? "middle" : side == 2 ? TOP   : BOTTOM;
+							   ori == 1 ? "middle" : side == 2 ? TOP   : BOTTOM;
```
leeoniya, line 1304: ...almost, it was all spaces before :)

joaopaulobdac: Here I go force-pushing again :rofl:

joaopaulobdac: Ok now it's all spaces.

leeoniya:
:shipit:

Issue #411, opened 2020-12-22
Question: Sync cursor and focus line programmatically


zefirka:
Hi @leeoniya!

I'm trying to write a tooltip plugin which renders tooltip on `setCursor` hook. Also I want to make tooltip synced between charts but only if I pressed Shift key. I can't figure out can I use native `cursor.sync` from uPlot or not with condition of pressed key. I though that I can do something like this: 

1. Create charts 
2. On Shift keypress update charts config to set `cursor.sync.key`
3. On Shift keyup update charts config to remove `cursor.sync.key`

Am I on the right way? 

And second question is about ability to focus line programmatically. Can I do that?

leeoniya:
right now there's no external way to disable sync programmatically, but i think it would be easy to add something like `u.setCursor({sync: {key: null}})`. internally it would simply call the same thing that `.destroy()` does:

https://github.com/leeoniya/uPlot/blob/d74bad7e15059c478928ce40d1059475acfffb59/src/uPlot.js#L2327

> And second question is about ability to focus line programmatically. Can I do that?

use `u.setSeries(1, {focus: true})` to focus and `u.setSeries(null, {focus: true})` to defocus all. note that to maintain explicit behavior, programmatic API calls will never sync -- it's assumed you will sync other charts via their APIs, if necessary.

bogdan-nourescu:
I suggest the second route, cause i recently hit a similar problem:
I want to make the mousewheel plugin to work on synced charts. So far i haven't found a way to make them work without getting hold of the sync.

leeoniya:
closing as dupe of #354

PR #413, opened 2020-12-25
Fixed typings for setSeries

zefirka:
Hi @leeoniya! 

You [said](https://github.com/leeoniya/uPlot/issues/411#issuecomment-749701974) in #411 that to defocus all lines it's should be called `setSeries(null, {focus: true})`, and I found that typing for `setSeries` not matches to that signature, so I've fixed it.


leeoniya:
thanks!

PR #421, opened 2020-12-28
Chartjs-next: Update for changed/optimal options

kurkle:


leeoniya:
hey @kurkle, sorry, i think i merged this prematurely.

the min/max optimizations is not something want to add. all other libraries scan the data to find the bounds. and many support zooming or panning naively, so these values cannot be assumed to be static as if this was a static chart.

PR #430, opened 2021-01-11
exposed pub/sub + added sync cursor programmatically demo

zefirka:
Hi @leeoniya in #411 yo [mentioned](https://github.com/leeoniya/uPlot/issues/411#issuecomment-749701974) that there is an option to expose the shared pub/sub globally. I've tried to solve my issue with that approach and it worked well so I've prepared this PR. 

Also I'm not sure about adding programmatically synced cursors in separate file. I put it `sync-cursor.html` firstly, but demo-page looked not too readable so I've decided to put it in separate file.


leeoniya:
thanks, i'll review this in a day or two. i'm finishing up some work for a 1.6 release :)

leeoniya:
i'm not a huge fan of how this is done, since it relies on poking the shared registry through the instances rather than just exposing it directly. if you look at my original suggestion, it was for exposing it as a static thing with methods that accept instances. it will be a bit more code, but i think would be more direct and more flexible. the fact that `u.pub()` exists to receive dispatched events should be an implementation detail and not an official public API.

simply exposing https://github.com/leeoniya/uPlot/blob/master/src/sync.js#L1 as `uPlot.syncs` should probably work and not require any pub/sub ops to be done via additional instance methods?

dist/uPlot.d.ts
```diff
@@ -164,6 +165,12 @@ type ArcV    = (p: Path2D, y: number, x: number, r: number, startAngle: number,
 type BezierCurveToH = (p: Path2D, bp1x: number, bp1y: number, bp2x: number, bp2y: number, p2x: number, p2y: number) => void;
 type BezierCurveToV = (p: Path2D, bp1y: number, bp1x: number, bp2y: number, bp2x: number, p2y: number, p2x: number) => void;
 
+type Sync = {
+	sub: (client: uPlot) => void;
+	unsub: (client: uPlot) => void;
+	pub: (type: string, client: uPlot, x: unknown, y: unknown, w: unknown, h: unknown, i: unknown) => void;
```
leeoniya, line 171: these are all numbers.

they're event properties in the context of the emitting instance (client) and are similar to `e.clientX` of DOM events. `x` & `y` are left/top coordinates, `w` & `h` are the css pixel dimensions of the plotting area, `i` is the hovered index. not all events will have all arguments set, some args may be null.

zefirka:
Sorry, I'm not sure that I figured out which way allow me to enable or disable sync between group of chart via static `syncs`. The one option in my mind is to make: 

```js
const a = new uPlot({cursor: {key: 'A', ....});
const b = new uPlot({cursor: {key: 'B', ....});

// and in sync switcher do something like 

uPlot.syncs['A'].sub(b);
uPlot.syncs['B'].sub(A);

// insted of 

a.sync.sub(b);
b.sync.sub(A);
```

I find this way more controllable, so if it is what you suggest - then I can fix this PR.

zefirka:
Also I noticed that when I'm syncing a couple of plots, and then disabling sync when cursor was set up on some point, then after it sync disabling it continues stands over that points, so I've thought that should be a convenient way to programmatically remove cusror, but I didn't find it, but found a workaround `setCursor({left: -1, top: -1})`. 

What do you think to add `u.setCursor(null)` which just removes cursor from plot?

zefirka:
Hi @leeoniya, can you look at my [question in this PR}(https://github.com/leeoniya/uPlot/pull/430#issuecomment-771052090)? 

Am I right in guessing what changes should I update?

leeoniya:
let me try a few variants today and see what feels best...

leeoniya:
took a slightly different approach, have a look: https://leeoniya.github.io/uPlot/demos/sync-cursor.html

Issue #429, opened 2021-01-11
Suggestion: allow to set custom focus colors for series


zefirka:
Hi! What do you think about to allow to set series focus color? Currently focus behavior just change alpha of series color, but if I want to make focus behavior a little different (e.g. B&W vs Colored, or a bit darken unfocused lines) I should change colors of series manually and call `.redraw()`. I suggest just add property `focusStroke/focusFill/...` for series for custom focus colors.

leeoniya:
> Currently focus behavior just change alpha of series color

if you noticed, alpha is actually applied to all _non-focused_ series. are you asking to add more overrides to non-focused series, or also the focused series? cause they would have to be different sets of options.

it looks like you're asking to add overrides to each series. i'm not a big fan of duplicating all the styling opts just to have customized focus behvior.

i'm open to adding a callback or two that makes it easier to add these yourself in a pluhin, though.

> but if I want to make focus behavior a little different (e.g. B&W vs Colored, or a bit darken unfocused lines)

i would like to add defocus styling for more than alpha, but sadly, Safari is new IE: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/filter

zefirka:
>if you noticed, alpha is actually applied to all non-focused series.

Yep, I've mean custom options for blurred lines too. 

> i'm open to adding a callback or two that makes it easier to add these yourself in a pluhin, though.

Sure, it would be very useful.

leeoniya:
here's the path of least resistance, i think:

add a `._focus` property to each series that gets toggled here:

https://github.com/leeoniya/uPlot/blob/e870fc16cef352fb283a050957ec08374e21a47f/src/uPlot.js#L1854-L1856

then each series can implement dynamic styling by checking that property.

e.g.

https://github.com/leeoniya/uPlot/blob/e870fc16cef352fb283a050957ec08374e21a47f/demos/gradient-fill.html#L73-L81

can you try this route and see how it works out?

leeoniya:
the only tricky one is dynamic lineWidth, but that's a deep rabbithole: https://github.com/leeoniya/uPlot/issues/403

zefirka:
Hi, I'm a little confused about place to hold focus callback. Do you suggest to implement something like this: 

```js
series: [
  {},
  {
    label: "Serie",
    stroke: "green",
    focus: (u, seriesIdx) => {
        let s = u.series[seriesIdx];

        if (s._focus) {
[...]

PR #440, opened 2021-01-25
Added series._focus property

zefirka:
Hi @leeoniya 

After discussions in #429 I've decided to implement `s._focus` approach to add ability for dynamic focus color.

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -759,6 +759,7 @@ export default function uPlot(opts, data, then) {
 			s.stroke = fnOrSelf(s.stroke || hexBlack);
 			s.fill   = fnOrSelf(s.fill || null);
 			s._stroke = s._fill = s._paths = null;
+			s._focus = true;
 
 			let _ptDia = ptDia(s.width, 1);
 			let points = s.points = assign({}, {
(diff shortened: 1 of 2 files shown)


src/uPlot.js
```diff
@@ -759,6 +759,7 @@ export default function uPlot(opts, data, then) {
 			s.stroke = fnOrSelf(s.stroke || hexBlack);
 			s.fill   = fnOrSelf(s.fill || null);
 			s._stroke = s._fill = s._paths = null;
+			s._focus = true;
```
leeoniya, line 762: hmm, this gets the job done but i'm not excited by the semantics :(

there are additional uses for indicating explicit focus of one or multiple series besides color modification.

what do you think about making this a 3-state property: `null`, `true`, `false`. when nothing is explicitly focused it will be `null`. but when one or more series are explicitly focused, they will be `true` and the rest will be `false`.

zefirka: Yeah, i was thinking about that but was not sure how exactly to display 3 possible states. I'll update this.

leeoniya:
:+1:

merged with a rebase

PR #445, opened 2021-02-03
Group typings into a namespace

### LINKED ISSUE #444: Bring back the uPlot typescript namespace
Hey @leeoniya :wave:  Since [this commit](https://github.com/leeoniya/uPlot/commit/a0fc0a8b62a39b81dd3b9eaafd66d40f983fbab2) from v1.4.0, all the types/interfaces related to uPlot are no longer in a namespace:

## Before
```ts
import uPlot from 'uplot'

const axisStyles: uPlot.Axis = {
    font: `500 12px ${htmlStyles.fontFamily}`,
    stroke: htmlStyles.color,
    grid: { stroke: 'rgba(0,0,0, 0.05)' },
    ticks: { show: false }
}

const plot = new uPlot()
```

## After
```ts
 // Loss of information about where the Axis type comes from.
import uPlot from 'uplot'
import type { Axis } from 'uplot'

const axisStyles: Axis = {
}

const plot = new uPlot()
```
or
```ts
// Having to import uplot twice.
import uPlot from 'uplot'
import type * as uPlotTypes from 'uplot'

const axisStyles: uPlotTypes.Axis = {
}

const plot = new uPlot()
```
or
```ts
// This is the closest to where we had before, but having to access the
// `default` export for the constructor is not good.
import * as uPlot from 'uplot'

const axisStyles: uPlot.Axis = {
}

const plot = new uPlot.default()
```

With a namespace that is named the same as the uPlot class like it was before, we can use the TypeScript and JavaScript values at the same time.

joaopaulobdac:
(closes #444)

### DIFF
--- a/dist/uPlot.d.ts
+++ b/dist/uPlot.d.ts
@@ -1,8 +1,8 @@
 declare class uPlot {
 	/** when passing a function for @targ, call init() after attaching self.root to the DOM */
 	constructor(
-		opts: Options,
-		data?: Data,
+		opts: uPlot.Options,
+		data?: uPlot.AlignedData,
 		targ?: HTMLElement | ((self: uPlot, init: Function) => void)
 	);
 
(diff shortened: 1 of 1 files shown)


leeoniya:
:+1: thanks for your work

PR #476, opened 2021-03-22
roundX -> toFixed(X)

mixartemev:
We don't need redundant functions `round2` and `round3`.


leeoniya:
they are there for performance. toFixed() produces a string, which is then convereted back to a number when doing math with it, while round() stays in number-land.

have you measured the perf impact of switching?

mixartemev:
Oh, it's clear. In that case roundX func is a good option indeed.

PR #483, opened 2021-04-03
Do not modify the data passed by client applications

skalinichev:
It's not a very good manners anyway, and it also triggers asserts
in React/Vue.js environments, since this action modifies component props.


leeoniya:
:+1: thanks!

btw, feel free to PR https://github.com/skalinichev/uplot-wrappers into https://github.com/leeoniya/uPlot#third-party-integrations ;)

PR #490, opened 2021-04-10
fixed drawSeries hook typings

zefirka:
According to [this code](https://github.com/leeoniya/uPlot/blob/master/src/uPlot.js#L1304) `seriesKey` is just series' index and it should be `number`.


leeoniya:
oops :)

👍

Issue #465, opened 2021-03-03
Ranger out of boundaries


damianoporta:
Hello 
I need to create a chart like this one: https://leeoniya.github.io/uPlot/demos/zoom-ranger-grips.html
I think it has a bug because the ranger does not stay properly within its boundaries.

Do we have a correct demo for it? I need to create a candlestick chart with a ranger like that.
Thanks

leeoniya:
> I think it has a bug because the ranger does not stay properly within its boundaries.

yes, there is some clamping missing to limit the drag ranges. i won't be able to get to fixing it till late next week. if you'd like to submit a PR with a fix, i will review it.

damianoporta:
@leeoniya I think we just need to fix the update function (https://github.com/leeoniya/uPlot/blob/master/demos/zoom-ranger-grips.html#L102-L105). checking:

1. if newLft is < 0 
2. if newLft + newWid > uRanger.bbox.width

is `uRanger.bbox.width` the correct wait to retrieve the width of the ranger?
Inspecting the uRanger of the demo, the width is 725px, is that correct?

leeoniya:
yep, fixed.

PR #491, opened 2021-04-11
Bug-fix/ranger out of boundary

eveshi:
To resolve issue #465


leeoniya:
hi @eveshi, thanks for the PR :)

there are a couple concerns here.

1. this does fix the drag, but does not fix the resize.
2. `update()` is called very frequently, so it needs to be very fast. running `getElementsByClassName()` and also reading `clientWidth` in this hot function will not be the best performance.

grips

i think the proper fix for this looks like this:

```js
function update(newLft, newWid) {
	let newRgt = newLft + newWid;
	let maxRgt = uRanger.bbox.width / devicePixelRatio;

	if (newLft >= 0 && newRgt <= maxRgt) {
		select(newLft, newWid);
		zoom(newLft, newWid);
	}
}
```

eveshi:
Thank you for your reply! Learned a lot!

PR #493, opened 2021-04-12
Fix table alignment and remove Chart.js 2.9.4

benmccann:
I noticed both the Chart.js rows weren't aligned with the others. Also, I figured there wasn't any reason to keep 2.9.4 since none of the other libraries have multiple versions and we wouldn't recommend anyone newly adopting Chart.js to use that version


leeoniya:
i'm not sure i agree with removing 2.9.3/2.9.4 that still have something like 95% of the active install base (see npm by-version stats tab), especially for a market leader like Chart.js. the purpose of the benchmark (and uPlot in general) is to encourage people to switch to faster libs, like Chart.js v3, and if all they see is that "Chart.js is already fast", they will never bother to update their dependencies. i should probably add back the pre-5.0 version of ECharts for the same purpose, but their market penetration is 25% of Chart.js.

ideally there would be a "Current aligned" and "Usage relative" variant of this benchmark like caniuse.com has, but alas no one has time to manually do or maintain all that testing.

benmccann:
Ok. I sent https://github.com/leeoniya/uPlot/pull/494 instead

PR #494, opened 2021-04-12
Fix alignment in benchmarking table on README

benmccann:


leeoniya:
new benchmarks are posted.

PR #500, opened 2021-04-23
Initialize series/axis after uPlot instance is fully constructed

skalinichev:
Otherwise uPlot instance will be missing several instance properties inside callbacks upon series initialization, but not upon further usage (addSeries), which is a bit confusing.

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -877,8 +877,6 @@ export default function uPlot(opts, data, then) {
 
 	self.delSeries = delSeries;
 
-	series.forEach(initSeries);
-
 	const sidesWithAxes = [false, false, false, false];
 
 	function initAxis(axis, i) {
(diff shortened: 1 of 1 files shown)


leeoniya:
thanks, i'll review later today.

i assume this is the context: https://github.com/skalinichev/uplot-wrappers/issues/1 (please link any relevant info next time)

skalinichev:
Yes, that's the one. Will try to provide all the relevant information in the future, sorry for the misunderstanding.

leeoniya:
merged. thanks :+1:

leeoniya:
btw, i haven't tested whether this is sufficient to avoid additional `.valToPos()` errors. even though this method will now exist, it still relies on the scale ranges to be initialized, which in an autoscaling scenario doesnt happen until after init()->setData() is done.

PR #502, opened 2021-04-24
don't update the legend/cursor when resizing, close #499

### LINKED ISSUE #499: _setSize trigger a setLegend event throught updateCursor during _commit
Hey there !

I'm not sure if this is an issue, but I noticed that the legend get reseted each time a `setSize` is called on the graph (for responsive by example). As I'm not sure what is the intended way for this, I report it as an "issue". 

My use case is that the legend will stick to the lastest value when not hovered. Like shown here :
Screenshot 2021-04-22 223905

However when resizing (or scrolling on mobile), the legend got reseted each time which cause some visual artifact.
This behavior is fixed on my side during test by setting `shouldSetLegend` to false in `_setSize` and adding a check in the updateCursor function (here https://github.com/leeoniya/uPlot/blob/master/dist/uPlot.esm.js#L4055) :
`if (idxChanged && shouldSetLegend) {
			legend.idx = idx;
			setLegend();
		}`

The code logic is :
`_setSize => commit => _commit => updateCursor => setLegend`

Thanks for your time and this pretty cool lib 💯

Martichou:
I tested this change proposed by @leeoniya in #499 on both mobile device and desktop.
The issue is fixed and does not seems to introduce any regression.


leeoniya:
great, thanks!

Issue #474, opened 2021-03-20
Time format depending on the locale


denyaalt:
Hi,
How to display time in locale specific format (bcp47 language tag)?

```
console.log(date.toLocaleDateString('en-GB')); // → "20/12/2012"
console.log(date.toLocaleDateString('ko-KR')); // → "2012. 12. 20."
```
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString](url)

Thanks.

leeoniya:
there are a lot of possibilities here, and some will work while others won't.

here is my initial attempt: https://jsfiddle.net/vhq7dgew/1/

```js
// cache the Intl formatter
let fmtOpts = {timeZone:'UTC',year:'numeric',month:'numeric',day:'numeric',hour:'numeric',minute:'numeric',second:'numeric'};
let format = new Intl.DateTimeFormat('ko-KR', fmtOpts).format;

function utcDate(ts) {
  return uPlot.tzDate(new Date(ts * 1e3), 'Etc/UTC');
}

let opts = {
  title: "Korean / UTC",
  tzDate: utcDate,
  width: 640,
  height: 280,
  axes: [
    {
      // axis ticks formatter
      values: (u, splits) => splits.map(ts => format(utcDate(ts)))
    }
  ],
  series: [
    {
      // legend value formatter (doesnt get proper UTC timezone)
      value: (u, ts) => format(utcDate(ts)),

      // uPlot's formatter works okay:
  //  value: "{YYYY}-{MM}-{DD} {HH}"
    },
    {
      stroke: "red",
    }
  ],
};

let data = [
  [1609459200, 1609480800],
  [        35,         71],
];

let uplot = new uPlot(opts, data, document.body);
```

unfortunately, it looks like the browser's built-in Intl formatter doesnt properly pick up the timezone-shifted date. uPlot's internal formatter does work, however and you can use `uPlot.fmtDate()` to produce correct output for the indicated timezone. however, there is more involved here. first, the axis tick formatters have many different formats depending on zoom level and you would have to provide a config array to properly format the axis ticks how you need:

https://github.com/leeoniya/uPlot/blob/1692dc3556a9b8c883abf8488859f4272ceb6ee4/src/opts.js#L137-L150

you would also have to provide the korean weekeday and month names (short and long). you can see how it's done here: https://github.com/leeoniya/uPlot/blob/master/demos/months-ru.html

[...]

PR #517, opened 2021-05-19
WIP: Rework timeStuffs to allow time formats to be specified

bytesnz:
I was wanting to be able to format dates differently from the default american style. After having a look into the code, I decided to:
- Rework *opts.js* `genTimeStuffs()` to allow formats to be passed as an opts object
- Change uPlot to generate time axes functions on initialisation.

It allows users to give the time component formats in a `timeFormat` object. For example, to use DD/MM/YYYY.
```js
                        ...
			const opts = {
				width: 1920,
				height: 600,
				title: "Super Data",
				timeFormats: {
				    md: "{D}/{M}"
				}
				series: [
					{},
					{
						stroke: "red",
					},
				]
			};

			let u = new uPlot(opts, data, document.body);
```

To use YYYY-MM-DD
```js
    ...
    timeFormats: {
        md: "{M}-{D}",
        MLmdyy: "{YYYY}-{M}-{D}"
    }
````

If not specified in `timeFormats` the original defaults will be used.

This is a bit more basic than the request in https://github.com/leeoniya/uPlot/issues/474, but it is a start for allowing different formats (I currently don't have time to dig into the Intl date format stuff atm).

One thing I was wondering about was also allow you to specify the months and days in the `timeFormat` object. In the example case of [Russian names](https://leeoniya.github.io/uPlot/demos/months-ru.html), this could mean you don't need to call `uPlot.fmtDate` and just give it the arrays in the `timeFormat` object

### DIFF
--- a/src/opts.js
+++ b/src/opts.js
@@ -58,23 +58,24 @@ export const wholeIncrs = oneIncrs.filter(onlyWhole);
 
 export const numIncrs = decIncrs.concat(oneIncrs);
 
-const NL = "\n";
-
-const yyyy    = "{YYYY}";
-const NLyyyy  = NL + yyyy;
-const md      = "{M}/{D}";
-const NLmd    = NL + md;
-const NLmdyy  = NLmd + "/{YY}";
-
-const aa      = "{aa}";
-const hmm     = "{h}:{mm}";
-const hmmaa   = hmm + aa;
-const NLhmmaa = NL + hmmaa;
-const ss      = ":{ss}";
+export const NL = "\n";
 
 const _ = null;
 
-function genTimeStuffs(ms) {
+export function genTimeStuffs(ms, opts) {
+	opts = opts || {};
+	const yyyy    = opts.YYYY || "{YYYY}";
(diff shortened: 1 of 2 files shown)


leeoniya:
hi @bytesnz , thanks for the PR.

i don't think this is a good idea, since it is already possible to specify the entire time formatting table, which gives you 100% control. i would like to avoid offering simplified/partial versions of this, since this can easily be accomplished with external helper functions that assemble this table. is there a specific format you cannot modify with the existing api?

bytesnz:
Ah. Can you? I  couldn't see a way of doing it when I looked at the code.. (looking at https://github.com/leeoniya/uPlot/blob/master/src/uPlot.js#L463 and https://github.com/leeoniya/uPlot/blob/095e103e7c1c6c07dbfb202708329caadde9f7e9/src/opts.js#L156)?

leeoniya:
`axis.values` can redefine the whole formatting table: https://github.com/leeoniya/uPlot/tree/master/docs#axis--grid-opts

if you want only a single, unconditional tick format, `axis.values` can also be a template string.

PR #519, opened 2021-05-24
Fix syncOpts assign in case of series.length == 1

zefirka:
Hi @leeoniya!

[This commit](https://github.com/leeoniya/uPlot/commit/5ce1bec45d08763014bc751347a67e224c6f9bf9) brings non-backward compatibility behavior with charts when there is no series but timeline ([see this](https://jsfiddle.net/bgptvj5f/)). 

Can you take this PR and release patch version?

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -2653,7 +2653,7 @@ export default function uPlot(opts, data, then) {
 			pub: retTrue,
 			sub: retTrue,
 		},
-		scales: [xScaleKey, series[1].scale],
+		scales: [xScaleKey, series[1] ? series[1].scale : null],
 		match: [retTrue, retTrue],
 		values: [null, null],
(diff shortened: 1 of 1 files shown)


leeoniya:
yeah, i think we should fix this for now.

generally, i'm hoping to move away from sync-by-position, (which `null` currently implies), unless you'd like to talk me out of it with a convincing case.

since all scales must be pre-defined during init, would it make sense to do something like `series[1] ? series[1].scale : Object.keys(scales).find(key => key != xScaleKey)`

leeoniya:
merged this as-is. will consider other options in future.

thanks!

zefirka:
Thanks, can you publish new patch release?

leeoniya:
yes, will be tagged later today.

PR #521, opened 2021-05-29
demos: stacked-series: optimize time taken in stack function by up to 20%

Martichou:
Hey! 

I needed a stacked chart for one of my project (even if I know this is bad) and ended up rewriting some part of the stack function used to optimize it.
I ended with a 20% improvement so I'm sending it back to you.
It's up to you to merge it or not ;) 

Thanks for the library anyway !

ps: in the `bands.push()`, `fill: () => null,` is not needed as null is the default comportement for bands.

### DIFF
--- a/demos/stacked-series.html
+++ b/demos/stacked-series.html
@@ -24,29 +24,36 @@
 		<p>The two charts below <strong>show exactly the same data</strong> (did you notice?)</p>
 		<script>
 			function stack(data, omit) {
-				let data2 = [];
-				let accum = Array(data[0].length).fill(0);
-				let bands = [];
+		            let d0lenght = data[0].length;
+		            let accum = Array(d0lenght);
+		            // Faster than .fill(0)
+		            for (let i = 0; i < d0lenght; ++i) accum[i] = 0;
+		            // Create a copy of the base x axis (data[0]) in a new array
+		            let stacked_data = [data[0]];
+		            let length = data.length;
+		            let bands = [];
 
-				for (let i = 1; i < data.length; i++)
-					data2.push(omit(i) ? data[i] : data[i].map((v, i) => (accum[i] += +v)));
+		            for (let i = 1; i < length; i++) {
+		                stacked_data.push(omit(i) ? data[i] : data[i].map((v, i) => (accum[i] += +v)));
+		            }
 
-				for (let i = 1; i < data.length; i++)
-					!omit(i) && bands.push({
-						series: [
-							data.findIndex((s, j) => j > i && !omit(j)),
-							i,
-						],
-						fill: () => null,
-					});
+		            for (let i = 1; i < length; i++) {
+		                !omit(i) && bands.push({
+		                    series: [
+		                        data.findIndex((_s, j) => j > i && !omit(j)),
+		                        i,
+		                    ]
+		                });
+		            }
 
-				bands = bands.filter(b => b.series[1] > -1);
+		            // Assure no value below -1 exists in our array
+		            bands = bands.filter(b => b.series[1] > -1);
 
-				return {
-					data: [data[0]].concat(data2),
-					bands,
-				};
-			}
+		            return {
+		                data: stacked_data,
+		                bands,
+		            };
+		        }
 
 			let xs = [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29,30],
 				d1 = xs.map((t, i) => i < 15 ? 20 + i : i == 15 ? 15 : i == 16 ? 14 : i == 24 ? 20 : i == 28 ? 15 : i == 29 ? 0 : 30),


demos/stacked-series.html
```diff
@@ -24,29 +24,36 @@
...
-				bands = bands.filter(b => b.series[1] > -1);
-
-				return {
-					data: [data[0]].concat(data2),
-					bands,
-				};
-			}
+		            let d0lenght = data[0].length;
+		            let accum = Array(d0lenght);
+		            // Faster than .fill(0)
```
leeoniya, line 29: this is...surprising. will have to check it out.

image

leeoniya:
thanks, will take a look in a bit

demos/stacked-series.html
leeoniya: in isolation they seem to be pretty much identical +/- noise.

for me (on arch linux, node v16.2) this runs in 6.2ms:

```js
let length = 1e6;

console.time("loop");
let accum = Array(length);
for (let i = 0; i < length; ++i)
  accum[i] = 0;
console.timeEnd("loop");

console.time("fill");
let accum2 = Array(length).fill(0);
console.timeEnd("fill");
```

an you give me a snippet where this is faster?

leeoniya:
@Martichou can you post a bench that shows this perf increase? there is certainly not enough data in this demo to show such a difference. also your hardware, OS, browser specs would help.

demos/stacked-series.html
Martichou: Yeah sure !
For simplicity let me link you to a github repo I just created for this.
You can see the code used, as well as the benchmarks and how I run it.

My result of that said repo using `node main.js` :
```bash
➜  JsBench node -v
v15.8.0
➜  JsBench node main.js
Stack x 163 ops/sec ±10.42% (80 runs sampled)
Stack_pre_opt x 162 ops/sec ±10.36% (82 runs sampled)
Stack_opt x 300 ops/sec ±0.51% (88 runs sampled)
Fastest is Stack_opt
```
The only difference between Stack_pre_opt and Stack_opt is that `.fill(0)`.

Here is the link to the said repo: https://github.com/Martichou/jsbench

Martichou:
Yes !

I answered your question for the bench on your review ;) 
If you need anything else let me know.

CPU: I7-8750H
OS: WIN-10 PRO
BROWSER: Google Chrome lastest stable

demos/stacked-series.html
Martichou: It can even be a x2 in speed on big data assets.

leeoniya: jebus, that's absurd. what possible benefit is there for it to be backed by JS still?:

https://stackoverflow.com/questions/48836753/why-does-array-prototype-fill-have-such-a-large-performance-difference-compare

i have a bunch of other stuff where i use `.fill()` "for performance", assuming it was most optimized :angry:

Martichou: (This is pure speculation)
From what I saw and undertsand, that can be because `.fill()` is coded in C++.
Where the for loop can take advantage of Javascript V8 optimization caching or something like that.

https://stackoverflow.com/questions/48836753/why-does-array-prototype-fill-have-such-a-large-performance-difference-compare

Martichou: But might be worth checking of replacing `.fill()` with for loops would be benefical across the whole lib :)

leeoniya: yeah, i'm checking some now. creating a helper function to do this fill op doesnt seem beneficial in a local scatter demo i have that fills 5 series with x/y pairs.

i think it's highly dependent on the full context and surrounding code. i've certainly seen identical functions get optimized differently in different contexts.

before:

```js
function randInt(min, max) {
	return Math.floor(Math.random() * (max - min + 1)) + min;
}

let points = 10000;
let series = 5;

let data = Array(series).fill(null).map(v => [
	Array(points).fill(null).map(v => randInt(0,500)),
	Array(points).fill(null).map(v => randInt(0,500)),
]);
```

after:

```js
function randInt(min, max) {
	return Math.floor(Math.random() * (max - min + 1)) + min;
}

function filledArr(len, val) {
	let arr = Array(len);

	if (typeof val == "function") {
		for (let i = 0; i < len; ++i)
			arr[i] = val(i);
	}
	else {
		for (let i = 0; i < len; ++i)
			arr[i] = val;
	}

	return arr;
}

let points = 10000;
let series = 5;

console.time("prep");
let data = filledArr(series, v => [
	filledArr(points, i => randInt(0,500)),
	filledArr(points, i => randInt(0,500)),
]);
console.timeEnd("prep");
```

leeoniya: with 10k points, prep0 is fastest. with 100k points, prep2 is:

```js
function randInt(min, max) {
	return Math.floor(Math.random() * (max - min + 1)) + min;
}

function filledArr(len, val) {
	let arr = Array(len);

	if (typeof val == "function") {
		for (let i = 0; i < len; ++i)
			arr[i] = val(i);
	}
	else {
		for (let i = 0; i < len; ++i)
			arr[i] = val;
	}

	return arr;
}

let points = 10000;
let series = 5;

console.time("prep0");
Array(series).fill(null).map(v => [
	Array(points).fill(null).map(v => randInt(0,500)),
	Array(points).fill(null).map(v => randInt(0,500)),
]);
console.timeEnd("prep0");

console.time("prep1");
filledArr(series, null).map(v => [
	filledArr(points, null).map(i => randInt(0,500)),
	filledArr(points, null).map(i => randInt(0,500)),
]);
console.timeEnd("prep1");

console.time("prep2");
filledArr(series, v => [
	filledArr(points, i => randInt(0,500)),
	filledArr(points, i => randInt(0,500)),
]);
console.timeEnd("prep2");
```

leeoniya: though this might be because prep2 benefits from the JIT pass during prep1 :man_shrugging:

Martichou: Yes that what I had in my mind, it depend on the context.
I'm not a pro in Javascript but do you think it would be possible to benchmark each independ function in the lib (that would use .`fill()`) in order to compare if an optimization would be benefical ?
I'm coding in Rust mainly and we can easily do that by exporting the function and use it in a bench (https://github.com/Martichou/sys_metrics/blob/master/benches/cpu_bench.rs#L3). Would be dope to have something similar.

leeoniya: i made another copy of the `filledArr` function and renamed it, then copy-pasted the same set of tests 4x.

so at least for the current V8 it looks like prep2 is always fastest for 100k-length arrays and is also fastest on 10k arrays on passes 2+. given that the difference where prep2 is slower are +2ms and where it is faster is -110ms (1e6 length), i think it makes sense to stop using `.fill()` entirely.

Martichou: From my testing using your code:
```
fill x 6,354 ops/sec ±2.92% (85 runs sampled)
prep2 x 9,424 ops/sec ±1.98% (90 runs sampled)
prep1 x 8,181 ops/sec ±2.95% (88 runs sampled)
Fastest is prep2
```
JIT doesn't come into play with this, prep2 is faster.

leeoniya: for a more simple variant, `.fill(0)` stays faster longer.

filledArr3() only starts to win out at around 1e6:

```js
console.time("prep0f");
Array(points).fill(0);
console.timeEnd("prep0f");

console.time("prep2f");
filledArr3(points, 0);
console.timeEnd("prep2f");
```

at 1e6, the difference is 1-2ms. at 10k `.fill(0)` is .04ms vs 0.4ms for filledArr3. i think the difference at the cheap fill levels is probably due to the extra `typeof` check.

Martichou: Can you share your `filledArr3` ? 
Because on my side I get completely different result (filledArr is constantly faster, for any size).

Martichou: Here are my results :
```js
➜  JsBench git:(master) ✗ node main.js
DATA_SIZE 10
fill x 10,253,467 ops/sec ±7.18% (71 runs sampled)
filledArr x 57,077,166 ops/sec ±14.02% (69 runs sampled)
Fastest is filledArr
➜  JsBench git:(master) ✗ node main.js
DATA_SIZE 100
fill x 2,158,962 ops/sec ±4.73% (73 runs sampled)
filledArr x 6,950,915 ops/sec ±10.98% (73 runs sampled)
Fastest is filledArr
➜  JsBench git:(master) ✗ node main.js
DATA_SIZE 1000
fill x 266,720 ops/sec ±5.88% (79 runs sampled)
filledArr x 755,902 ops/sec ±8.22% (67 runs sampled)
Fastest is filledArr
➜  JsBench git:(master) ✗ node main.js
DATA_SIZE 100000
fill x 712 ops/sec ±21.59% (43 runs sampled)
filledArr x 698 ops/sec ±15.19% (39 runs sampled)
Fastest is filledArr,fill
➜  JsBench git:(master) ✗ node main.js
DATA_SIZE 1000000
fill x 129 ops/sec ±11.12% (56 runs sampled)
filledArr x 291 ops/sec ±11.70% (64 runs sampled)
Fastest is filledArr
```

Until `1e5` `filledArr` is still faster. At `1e5` there is a slight difference where `fill` would be faster.
At `1e6`, `filledArr` is faster again.

Note that my testing are done using multiple samples. So JIT can influence.

leeoniya: so, the only place that uses `Array.fill()` (besides demos) is in the `uPlot.join()` util function, which is not used by the core:

https://github.com/leeoniya/uPlot/blob/ff60e0f3fc11f135478ad22871128125e6ac77f2/src/utils.js#L405

i perf tested it against a plain `for` fill loop and didnt see a meaningful difference. you can use this demo to tweak the data sizes and watch the console: https://github.com/leeoniya/uPlot/blob/master/demos/align-data.html#L70

i'll leave this one as-is.

leeoniya:
i've merged the minimal changes to swap out `Array.fill()`. (this PR does too many unrelated changes).

btw, `fill: () => null` is still needed. without this re-stacking fails during toggle. try it here:

https://leeoniya.github.io/uPlot/demos/stacked-series.html

thanks for the discussion!

leeoniya:
actually, the reason `fill` was needed was because of a bug in the recently-introduced `.addBand()` method. during init the bands get this property auto-set here:

https://github.com/leeoniya/uPlot/blob/0edaa422d3f972dadcd615ff26363042682764fd/src/uPlot.js#L330-L332

and before the formal method was added, the bands were mutated/replaced directly, which required the fill to be present in the replacement bands.

this is fixed by https://github.com/leeoniya/uPlot/commit/0edaa422d3f972dadcd615ff26363042682764fd

good catch!

Martichou:
Nice ! I'm happy that my PR fixed at least one bug ;)

PR #533, opened 2021-06-17
Fix rangeNum type definition to make it callable

flyingmutant:
Before, it was impossible to call, as no arguments can satisfy
both possible types. Error when attempting to call e.g.

  `rangeNum(0, 0, 0, false);`

would look like

> TS2345: Argument of type 'number' is not assignable to parameter of type 'number & Config'.
>    Type 'number' is not assignable to type 'Config'.

### DIFF
--- a/dist/uPlot.d.ts
+++ b/dist/uPlot.d.ts
@@ -126,7 +126,7 @@ declare class uPlot {
 	static assign(targ: object, ...srcs: object[]): object;
 
 	/** re-ranges a given min/max by a multiple of the range's magnitude (used internally to expand/snap/pad numeric y scales) */
-	static rangeNum: ((min: number, max: number, mult: number, extra: boolean) => uPlot.Range.MinMax) | ((min: number, max: number, cfg: uPlot.Range.Config) => uPlot.Range.MinMax);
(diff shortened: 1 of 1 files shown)


leeoniya:
this is a tricky situation.

on one hand it fixes the TS issue, on the other hand it makes it difficult to understand the proper signature variations. after this PR it becomes unclear whether `extra` has any effect in the `uPlot.Range.Config` variant, and makes the argument name confusing where `mult` can be either a padding multiplier or a config.

unfortunately, typescript doesnt allow such a difference in signatures (and number of args) for function overloading :(

https://www.tutorialsteacher.com/typescript/function-overloading

i'm tempted to recommend using `// @ts-ignore` here, otherwise maintenance of this typings file will become nightmarish.

leeoniya:
great news! @joshhunt came to the rescue and figured out how to do this overload properly ^^

```js
import uPlot from 'uplot';

uPlot.rangeNum(0, 0, 0, true); // ok, proper typehints
uPlot.rangeNum(0, 0, { min: {}, max: {} });  // ok, proper typehints
uPlot.rangeNum(0, 0, { min: {}, max: {} }, false);  // error
```

please try it out

flyingmutant:
Works great, thanks! Did not know TypeScript allowed for these kinds of overloads, heh :-)

Issue #543, opened 2021-07-07
Suggestion: clip path in heatmap example


tmcnitt:
Great work on this library, it's by far the best I've found. On the [heatmap example](https://github.com/leeoniya/uPlot/blob/master/demos/latency-heatmap.html#L149), a path is used for clipping the drawn in rects. However when I tried to recreate it with an example that added data later, the x axis would not redraw correctly. When I save the context before creating the path, and restore it at the end of the plugin, everything works correctly. Could the example code be updated to better reflect that?

PR #544, opened 2021-07-07
Save and restore context in heatmap examples

tmcnitt:
Solves issue #543 
Can close if believe is a non-issue, but would have saved me some time

### DIFF
--- a/demos/latency-heatmap.html
+++ b/demos/latency-heatmap.html
@@ -145,7 +145,8 @@
 							const { ctx, data } = u;
 
 							let yData = data[3];
-
+							
+							ctx.save()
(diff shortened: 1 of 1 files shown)


leeoniya:
:+1: looks good, thanks!

tmcnitt:
It might be helpful to mention in the documentation that the context is mutable and not restored when a hook is called. It could be easy to make changes in the hook that mess up the context for uPlot

leeoniya:
`u.ctx` is exposed generally (not just used in hooks), and is a mutable web platform interface, so i think this can be assumed about it without special notes. there are certainly more important documentation holes than this!

PR #572, opened 2021-08-30
Use em-dash instead of double-hyphen in legend

flyingmutant:


leeoniya:
this is probably technically correct, though visually i still prefer `--`, already a lot of solid lines/borders in the legend.

will have to think about it :thinking:

PR #606, opened 2021-10-31
new candlestick chart demo

euprogramador:
implemented demo use horizontalPanZoomScrollerPlugin() function.

Features demonstrated:

- \+ / - keys to manage zoom
- use mouse whell to horizontal
- click and move to pan chart
- y rescale

### DIFF
--- a/demos/candlestick-ohlc-pan-zoom-scroller.html
+++ b/demos/candlestick-ohlc-pan-zoom-scroller.html
@@ -6,6 +6,7 @@
 	<title>Candlestick Chart</title>
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<link rel="stylesheet" href="../dist/uPlot.min.css">
+	
 </head>
 
 <body>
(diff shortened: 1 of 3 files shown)


leeoniya:
thanks @euprogramador , i'll try to review this soon.

there's a good chance i'll backport some of the improvements from https://github.com/grafana/grafana/pull/40909 to the current candlestick demo. generally i'd prefer to have fewer variants of the same code that can easily become out of sync, so it would be good to roll more stuff into the existing demo, but probably not before i update the original. WDYT?

demos/candlestick-ohlc-pan-zoom-scroller.html
```diff
@@ -163,6 +164,55 @@
 
 				}, false);
 
+
+			
+				/* left axis scale */
+
+				let leftAxisScale = u.root.querySelectorAll(".u-axis")[1];
```
leeoniya, line 171: great to see use of cutting-edge features :)

this was exactly the use case i had in mind for these rects, too.

euprogramador: I'm implementing a trading system, and the graphic visualization in metatrader doesn't support the graphic types I want to create. I will actually use context2d to create cluster footprint charts. vertical scale will be required to implement these features.

as I saw that others wanted this, I decided to implement it in the example to help those who also want these features, I thought your project was really cool and very fast.

leeoniya:
image

Issue #607, opened 2021-11-01
Demos broken in Safari 12 and 13 


samuel-git:
Great work with this library!

Older Safaris (and maybe other browsers) crash the javascript thread with ``TypeError: n.addEventListener is not a function``. Tested here: https://leeoniya.github.io/uPlot/bench/uPlot.html on Safari 12 MacOS Mojave and Safari 13 on MacOS Catalina.

This is the problematic line: https://github.com/leeoniya/uPlot/blob/cdc31b47b77fad5d0ad911e231b2892011b58e26/dist/uPlot.esm.js#L645

Called from here: https://github.com/leeoniya/uPlot/blob/cdc31b47b77fad5d0ad911e231b2892011b58e26/dist/uPlot.esm.js#L561
Without this line everything works fine, my guess is that the event listener on matchMedia(...) is not supported by older Safaris.
<img width="780" alt="grafik" src="https://user-images.githubusercontent.com/22152728/139668271-2fca33a7-3618-446c-adb0-50e8943a07b0.png">

Edit: Add screenshot and better error description

leeoniya:
please search for previous issues before opening new ones: https://github.com/leeoniya/uPlot/issues?q=is%3Aissue+safari

the fix is to polyfill older browsers if you need support for them: https://github.com/leeoniya/uPlot/issues/538#issuecomment-870711531

samuel-git:
Thanks for pointing me in the right direction and excuse my mistake. I would suggest writing something about polyfills in the readme when not included in the code as I am not the first person with this issue und these are relatively new Browsers which should support matchMedia() https://caniuse.com/matchmedia so this is not obvious.

PR #608, opened 2021-11-01
Fix crashes in older Safaris

samuel-git:
Should fix https://github.com/leeoniya/uPlot/issues/607

Not sure how important the event listener is on the matchMedia query, it does not get attached with these changes on unsupported browsers instead of crashing.


leeoniya:
see https://github.com/leeoniya/uPlot/issues/607#issuecomment-956249687

PR #611, opened 2021-11-03
Update typings for Sync.Scales to receive null

zoltanbedi:
Line 493 states that it is valid to pass scales like this [key, null] but the typing is not in sync with that.

### DIFF
--- a/dist/uPlot.d.ts
+++ b/dist/uPlot.d.ts
@@ -466,7 +466,7 @@ declare namespace uPlot {
 		}
 
 		export namespace Sync {
-			export type Scales = [xScaleKey: string, yScaleKey: string | null];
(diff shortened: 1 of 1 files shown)


dist/uPlot.d.ts
```diff
@@ -466,7 +466,7 @@ declare namespace uPlot {
 		}
 
 		export namespace Sync {
-			export type Scales = [xScaleKey: string, yScaleKey: string];
+			export type Scales = [xScaleKey: string, yScaleKey: string | null];
```
zoltanbedi, line 469:
```suggestion
			export type Scales = [xScaleKey: string | null, yScaleKey: string | null];
```

leeoniya:
nice, thanks! :+1:

Issue #538, opened 2021-06-23
uPlot giving error and not working below iOS version 13


abhisheknaiidu:
Currently, not working in iOS mobile version below 13 and showing this - 

Some `MediaQueryList` or `el.AddListner` is causing problem.

<img width="782" alt="Screenshot 2021-06-24 at 3 00 36 AM" src="https://user-images.githubusercontent.com/55599878/123170916-b57e4180-d498-11eb-9696-49c92011b9d2.png">

`After consoling`:

<img width="969" alt="Screenshot 2021-06-24 at 3 16 39 AM" src="https://user-images.githubusercontent.com/55599878/123172342-a4cecb00-d49a-11eb-9d6a-5aba12310a8a.png">

leeoniya:
also reported & explained here: https://github.com/grafana/grafana/issues/35594#issuecomment-864302938

long story short you can update safari to v14 or later, which has been available since Sept 2020 for all recent iOS and macOS versions.

or you can add some simple polyfill code that will solve this in your app like

```js
MediaQueryList.prototype.addEventListener = MediaQueryList.prototype.addEventListener || function(type, cb) { this.addListener(cb); }

MediaQueryList.prototype.removeEventListener = MediaQueryList.prototype.removeEventListener || function(type, cb) { this.removeListener(cb); }
```

(untested)

abhisheknaiidu:
@leeoniya you mean like this - 

```js
MediaQueryList.prototype.addEventListener = MediaQueryList.prototype.addEventListener ?? function(type, cb) { this.addListener(cb); }

MediaQueryList.prototype.removeEventListener = MediaQueryList.prototype.removeEventListener ?? function(type, cb) { this.removeListener(cb); }

function on(ev, el, cb, capt) {
	MediaQueryList.addEventListener(ev, cb, capt ? evOpts2 : evOpts);
}

function off(ev, el, cb, capt) {
	MediaQueryList.removeEventListener(ev, cb, capt ? evOpts2 : evOpts);
}
```

as el is MediaQueryList only

abhisheknaiidu:
Working by adding like this - 
```js
	console.log(el, ev)
	if (el.addEventListener) {
		el.addEventListener(ev, cb, capt ? evOpts2 : evOpts)
	  } else {
		el.addListener(cb)
	  }
```

Is there any way to add this without changing in the `\dist` file

leeoniya:
you should not need to patch on() and off().

[...]

PR #622, opened 2021-11-24
fix: old safari bug

wangtao0101:


wangtao0101:
@leeoniya can you release a verion soon for this bug

leeoniya:
i dont plan to support old browsers directly in uPlot. if you need old browser support, please use polyfills. in this case it's Safari < 14. a working shim is shown in https://github.com/leeoniya/uPlot/issues/538#issuecomment-870711531

thanks for thr PR anyhow :)

PR #649, opened 2022-01-10
add a null-check of `xKeySrc` in `cacheMosue`

FrodoLuo:
Since a `TypeError`(reading `ori` of `undefined`) will be thrown causing blocking in browser when `config.cursor.sync.scales` was given a `[null, any]`;

Added a null checking before using it as a key

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -2697,7 +2697,7 @@ export default function uPlot(opts, data, then) {
 			let [xKeySrc, yKeySrc] = syncOptsSrc.scales;
 			let [matchXKeys, matchYKeys] = syncOpts.match;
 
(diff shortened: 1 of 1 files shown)


leeoniya:
thanks for the PR :)

i guess this means it's not currently possible to sync x by position for rotated charts. i think it makes more sense to assert src orientation without relying on the x sync scale. something like

```js
let rotSrc = src.scales[src.axes[0].scale].ori == 1;
```

or maybe a bit more efficient/terse:

```js
let rotSrc = src.axes[0].side % 2 == 1;
```

this is essentially pretty how chart orientation inferred everywhere else.

FrodoLuo:
yes, in my case, we are trying to sync cursors between timeseries graphs by relative position.

and let me do some modification :)

PR #661, opened 2022-02-13
types: remove empty statement

Maxdamantus:
This was causing the TypeScript error, "Statements are not allowed in ambient contexts":

```
node_modules/uplot/dist/uPlot.d.ts:952:48 - error TS1036: Statements are not allowed in ambient contexts.

952   export interface Border extends OrthoLines {};
```


leeoniya:
thanks!

PR #673, opened 2022-03-18
demos/nice-scale: Adapt `devicePixelRatio`

### LINKED ISSUE #672: Nice scale not work.
### Reproduce demo
https://leeoniya.github.io/uPlot/demos/nice-scale.html

### What happend
there is no display value in yAxis.

### What expected
yAxis has "nice scale", it always cover the actual minmax data.

### Other
Sometimes the demo above could be re-setScale infinitly but i didn't know what reason cause it.

Fiv5:
fix #672

### DIFF
--- a/demos/nice-scale.html
+++ b/demos/nice-scale.html
@@ -89,7 +89,7 @@
 							yDataMin = dataMin;
 							yDataMax = dataMax;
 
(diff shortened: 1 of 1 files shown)


leeoniya:
nice, thanks :+1: 

i hope to improve scales soon (in 2.0) so the default behavior is closer to this demo.

PR #679, opened 2022-03-29
Check that matchMedia is available before running it

mharju:
Our Jest unit tests fail to run due to a missing `matchMedia` function. I think we could mock or make this work in our use case but this would be nice to be handled in the library itself.


leeoniya:
hey @mharju 

a Jest mock exists for this. i'd prefer not to gate features which are expected to exist in all browser envs.

https://www.npmjs.com/package/jest-matchmedia-mock

mharju:
> hey @mharju
> 
> a Jest mock exists for this. i'd prefer not to gate features which are expected to exist in all browser envs.

Sure, that what I was referring to – we can do that too and I understand your position.

Thanks for the amazing work you have done with uPlot, it's by far the best graphing library for our particular needs and is able to stretch to amazing heights with some custom plugins.

PR #688, opened 2022-04-19
make window optional on import library

jalopez:
## Problem
We want to do SSR of grafana/ui library which depends on this one. We only need that the library can be imported in server side (no actual usage)

## Solution
- Make all initialization code optional depending on the existence of `window`
- Remove `doc` and `win` variables as they were global

Context https://github.com/grafana/grafana/pull/46288


leeoniya:
hey @jalopez, thanks for the PR!

i made a branch based on this work but with fewer changes: https://github.com/leeoniya/uPlot/tree/ssr-env-noop. i omitted the `/dist` files for easier review.

can you plz try building and re-testing that branch? the following seems to work now without errors, so hopefully the ESM variant also works :crossed_fingers: 

```js
let uPlot = require('./uPlot.cjs.js');
```

leeoniya:
should be fixed on `master` now. thanks :+1:

PR #800, opened 2023-02-14
[docs] fix function signature for axis.space

grrtbrtr:
The documentation mentioned the following function signature for the `options.axes[x].space` property.
- Current entry: `(self, scaleMin, scaleMax, dim) => space`
- Corrected entry: `(self, axisIdx, scaleMin, scaleMax, dim) => space`

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -372,7 +372,7 @@ let opts = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


leeoniya:
thanks!

PR #811, opened 2023-03-21
Tweak best practices for LCJS bench

niiloArction:
1. Disable theme effects for performance benchmarks. This is enabled by default, but does not reflect best performance.
2. Fix incorrect syntax for line series data pattern.

I didn't find instructions on how the benchmark results were measured, so not sure how this would reflect on them. But with `requestAnimationFrame` load speed test the loading seems to be ~100 milliseconds faster.

We were a bit late on updating our own performance test bench code with the latest best practices, so its understandable that some of this was missed. Would appreciate updating the results but I totally understand if its not done immediately.

### DIFF
--- a/bench/LightningChart.html
+++ b/bench/LightningChart.html
@@ -48,7 +48,6 @@
 
 			function makeChart(data) {
 				console.time('chart');
-
 				const {
 					lightningChart,
 					SolidLine,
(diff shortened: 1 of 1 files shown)


leeoniya:
tested with your changes. the scripting time improvement is ~30ms (out of 250ms).

i normally do forced GC between continuous page reloads to purge JIT cache, force re-parsing of libs and simulate fresh boot. this affects large libraries much more than small ones. without forced GC, i get 140ms scripting for LightningChart but with forced GC i get 240ms scripting (with Evaluate Script taking ~85ms).

for reference, uPlot is ~40ms for both GC and non-GC runs, with Evaluate Script < 2ms.

i'll consider re-doing the bench without forced GC in the future, but i also i want to reward libs for striving for optimized bundle size, so it's not clear that this specific metric should be changed to omit this cost.

**uPlot with GC**:

https://user-images.githubusercontent.com/43234/226749233-7b867182-9069-49e9-8243-ab3b4fc4782c.mp4

---

**uPlot no GC**:

https://user-images.githubusercontent.com/43234/226749240-3cd66458-66ef-4f3e-93ec-be5be4e20bbb.mp4

---

**LightningChart with GC**:

https://user-images.githubusercontent.com/43234/226749235-ea465458-4693-44e1-a7de-244d26d2717d.mp4

---

**LightningChart no GC**:

https://user-images.githubusercontent.com/43234/226749239-baa4d241-b2e0-4d5d-98dc-fa45651ec0cb.mp4

PR #828, opened 2023-05-04
Remove non-existing name `matchSeries`

KyleBsingh:
uPlot.d.ts is broken, as there is mention of a non-existing name


leeoniya:
oops, thanks!

PR #860, opened 2023-08-01
Initialize scales with range check

adela-almasan:
On initialization the scales are always initialized to the data range even when a range function exists.


leeoniya:
thanks for the PR :)

i solved it a bit differently: https://github.com/leeoniya/uPlot/commit/c749b61abe942a38d16110bf44ebf16689445280

that dataset it not really valid anyways and will have additional issues with hovering, stacking, and a bunch of other things. added a comment in https://github.com/grafana/support-escalations/issues/6671#issuecomment-1661376244

PR #900, opened 2023-12-11
Fix null error in syncLegend-method

progitect:
This pull request fixes a null error in the syncLegend function. This error would occur when syncLegend was called before the legends were fully loaded/created.

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -2394,8 +2394,10 @@ export default function uPlot(opts, data, then) {
 
 				let j = 0;
 
-				for (let k in vals)
-					legendCells[i][j++].firstChild.nodeValue = vals[k];
+				for (let k in vals) {
+          if (legendCells[i][j].firstChild?.nodeValue)
+            legendCells[i][j++].firstChild.nodeValue = vals[k];
+        }
 			}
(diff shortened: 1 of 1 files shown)


leeoniya:
thanks for the PR :)

> before the legends were fully loaded/created.

would you mind making a jsfiddle similar to https://jsfiddle.net/0bvks7cw/1/ that reproduces this issue?

i think there might be a better way to fix this but i'll need to verify that it works.

progitect:
Thanks for your input. I will try to reproduce it in a jsfiddle, but I do not have the time at this moment to create it. Will update when I have reproduced it.

progitect:
After trying to reproduce this issue in a jsfiddle, it turned out that the problem was only occurring when syncing to a timeline chart ([https://leeoniya.github.io/uPlot/demos/timeline-discrete.html](https://leeoniya.github.io/uPlot/demos/timeline-discrete.html)) like the 'Merged same consecutive states' one. I have fixed it with some extra code in the timeline plug-in. Basically I added some code to force that the `setCursor` hook is only executed when all series are drawn. 

As the isssue is solved I will close this PR,

PR #920, opened 2024-01-26
Corrected mouse position during focus if scaleX.ori is 1

alex-tel:


leeoniya:
thank you!

i fixed this by moving up the `mouseYPos` var from inside the `bias` block that's a bit lower.

also added some focus opts to the orientation demo: https://leeoniya.github.io/uPlot/demos/scales-dir-ori.html

alex-tel:
Could you please add this (in the red circle?) without it, vertical series is not focused (tested when each series has own scale) 
Screenshot 2024-01-26 at 17 52 30
Screenshot 2024-01-26 at 18 01 22
Screenshot 2024-01-26 at 18 01 17

leeoniya:
i added this changes in the linked commit above (https://github.com/leeoniya/uPlot/commit/baec062a89b9c13746e8cc602d420d6974da8fd9)

PR #985, opened 2024-08-11
Add in localization for datetime with override option for datetime and numbers

### LINKED ISSUE #474: Time format depending on the locale

willfarrell:
Features:
- new option to override browser locale for measurment and datetime
- timeAxisStamps updated to leverage Intl.DateTimeFormat provided by @Christian-Sander
- numbers and datetime will now show in the same locale now

Breaking changes:
- fmtDate using strings is no longer supported

Notes:
- type defs might need a little update. I don't know TS.
- Will be testing over the next week.

Closes #474


leeoniya:
hey @willfarrell

thanks for the PR :)

since this introduces many breaking changes, i will not be able to merge it into any v1.x release in its current form. perhaps it is possible to implement this as an additionally supported formatting strategy rather than breaking existing templates?

it changes the behavior of many demos, for example https://leeoniya.github.io/uPlot/demos/timezones-dst.html no longer shows hours/minutes, which is the main purpose of that demo.

also, i don't think this PR addresses the concern raised in #474 of being unable to insert line breaks into the tick strings?

anyways, this is a good starting point for something that might evolve into something real. :+1:

willfarrell:
Yes, this is a starting point, but wanted to make sure it was shared early. Sorry, should have converted to a draft.

Agreed, I used the `timeAxisStamps` provided by a commenter, looks like it needs some tweaking. Maybe another demo for all the stamp patterns would be helpful for testing this.

> perhaps it is possible to implement this as an additionally supported formatting strategy rather than breaking existing templates?
I looked into this initially, but couldn't figure out how to override the existing `timeAxisStamps` via options.

> unable to insert line breaks into the tick strings?
I'll reread in details, must have missed that. Was kinda focused on my own use case and saw it has a lot of overlap.

willfarrell:
PR is updated to support multi-line date/times and localized legend date-time. 

Notable differences (en-us) at lower zoom levels:
- 2-digit vs 4-digit year
- `,` added between date and year
- ` ` added between time and hourCycle indicator
These make the date-time 4-char longer, but I feel this is fine because it's on a different line.

Screenshot 2024-08-17 at 07 24 10

Did some testing with a bunch of locale, using https://en.wikipedia.org/wiki/List_of_date_formats_by_country seems to look good.

PR #987, opened 2024-08-15
Default to crosshair cursor in plot area

PabloAlexis611:
I think using a crosshair cursor style over the default pointer allows the user to better understand what data point they're hovering on. Makes it easier to connect x/y axis values visually in the plot .

### DIFF
--- a/src/uPlot.css
+++ b/src/uPlot.css
@@ -27,6 +27,10 @@
 	position: absolute;
 }
 
+.u-over {
+	cursor: crosshair;
+}
+
 .u-under {
 	overflow: hidden;
 }


leeoniya:
interesting. i'm not sure this is a good default since uPlot has its own crosshairs and cursors often are rendered with shadows and if the alignment isnt perfectly on-center it can be quite distracting:

image

PabloAlexis611:
Mmm, that's interesting - I had never noticed that slight off-center issue.

From my experience I found it more distracting to have a regular pointer cursor because I see a different shape that isn't "connected" in a way to what I'm trying to hover - the default uPlot crosshairs are great, but I liked the extra highlight on what I'm selecting with a crosshair cursor instead.

I probably didn't notice the offset/shadow since I set my background to a darker color.

PR #1022, opened 2024-12-11
Add side to grouped bar plugin, fix alignment with different sides

gelicia:
This does not fix every axis/alignment scenario

- Adds an example with axes on the top, right, bottom and left. Each axes example has an example with the text being left or right aligned
- adds an optional ability to align text in the grouped bar plugin
- Adds logic to uplot where if the orientation is horizontal, it will align text to 0 or the chart width when the text is aligned left or right.

Left set of examples is before the change, right is after the change. 
Screenshot 2024-12-11 at 9 22 49 AM

Related to https://github.com/grafana/grafana/issues/94344


dist/uPlot.iife.js
```diff
@@ -4564,8 +4564,13 @@ var uPlot = (function () {
 					if (val != null) {
 						if (ori == 0)
 							x = canOffs[i];
-						else
+						else {
+							if (axis.side === 1 && axis.align === 2)
+								x = can.width;
```
gelicia, line 4569: I'm guessing 0 to canvas width is not the correct boundaries we want - what are the best offsets?

leeoniya:
hey, thanks for this PR!

i took a slightly different approach in https://github.com/leeoniya/uPlot/commit/63a9b4a233fa2d4d2b334471fdb3487398a5e648

PR #1026, opened 2024-12-28
add demo showing value and line in axis overlay

RonnieLincoln:
image

### DIFF
--- a/demos/axis-with-value-and-line.html
+++ b/demos/axis-with-value-and-line.html
@@ -1,0 +1,208 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


leeoniya:
thanks so much! it's clear you did your homework, especially with using `cursor.bind` :)

i refactored things a bit and pulled it out into a plugin, and extracted some of the static styles to CSS. i understand why you made some of the choices that you did, and some of the things i changed (like not reading back from uPlot instance) may not be to everyone's liking, and in some ways makes the setup a bit more fragile, but feel free to incorporate any of the refactorings that you like into your own code.

https://leeoniya.github.io/uPlot/demos/axis-indicators.html

see you in 2025!

PR #1040, opened 2025-02-25
fix series toggle when `cursor.points.show()` may have returned `undefined`

Estasie:
After update from 1.6.27 to 1.6.31 faced with a problem of toggling series at the chart.
I give serie option show field the value of false or some function usually, and now error happens at 
Screenshot 2025-02-25 at 16 17 25
because el is undefined, to prevent this error would be good to check if pt is not undefined or null. Have done it at the scene, but can move directly to elTrans function. 

But it's better to search the root cause, but I gave up on fnOrSelf moment, so asking for a help or accepting this change.

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -2418,7 +2418,7 @@ export default function uPlot(opts, data, then) {
 
 			if (showCursor) {
 				let pt = cursorOnePt ? cursorPts[0] : cursorPts[i];
-				elTrans(pt, -10, -10, plotWidCss, plotHgtCss);
+				pt && elTrans(pt, -10, -10, plotWidCss, plotHgtCss);
 			}
 		}
 	}


leeoniya:
hey @Estasie,

this is the first such report, which i think is pretty odd. can you make a small example that reproduces your issue?

kelvin-chalk:
Hi @leeoniya, 

I am also experiencing this issue. This is a fairly recent regression introduced in [1.6.31](https://github.com/leeoniya/uPlot/compare/1.6.30...1.6.31#diff-6f581b44485db1123ab1cc1b28b05d08b22ddb46d7ad1d19252acb9f15debf3dR2415) 


In our chart implementation, the chart updates almost as expected but throws an error that prevents execution of hooks onto relevant functions (for us, `setSeries`). This reproduces when `setSeries` is called by a UI element that exists outside of the chart - i.e. when cursorPts would be undefined.

leeoniya:
> This reproduces when `setSeries` is called by a UI element that exists outside of the chart - i.e. when cursorPts would be undefined. 

i'm confused, are you calling `u.setSeries()` with an index outside of the series array?

can you please put together a simple jsFiddle that works in 1.6.30 but fails in 1.6.31? like https://jsfiddle.net/q2m7o1dc/

@Estasie would you please also make a jsFiddle that fails for you. what you provided above has way too much unrelated stuff going on. i need to be able to repro the issue easily without trying to re-assemble things from partial snippets.

leeoniya:
if we look at the initialization path for `cursorPts`:

https://github.com/leeoniya/uPlot/blob/b24436df67cfa393cc7e51a5b328288dec6cfe49/src/uPlot.js#L3454

if `cursor.show` is true, the point always gets initialized:

https://github.com/leeoniya/uPlot/blob/b24436df67cfa393cc7e51a5b328288dec6cfe49/src/uPlot.js#L1059

this `showCursor` condition is checked for inside of `toggleDOM` before interacting with `cursorPts`:

https://github.com/leeoniya/uPlot/blob/b24436df67cfa393cc7e51a5b328288dec6cfe49/src/uPlot.js#L2423-L2426

i'm struggling to understand what scenario would cause `cursorPts[i]` to be absent.

leeoniya:
okay, it can fail with something like this:

```
cursor: {
  points: {
    show: false,
  },
},
```

combined with

https://github.com/leeoniya/uPlot/blob/b24436df67cfa393cc7e51a5b328288dec6cfe49/src/uPlot.js#L989-L1000

this will return `undefined` cursor point elements.

leeoniya:
typescript would have helped here :sweat_smile:

kelvin-chalk:
Thanks! Sorry, I should have included a JSFiddle - we'll do that for future bug reports 😄

PR #1051, opened 2025-04-13
Logo for μPlot

MUmarShahbaz:
<div align="center"><br><h1>μPlot</h1></div>


leeoniya:
hey @MUmarShahbaz , thanks for this!

i actually had a similar idea, but using μ as the trendline itself. it would go negative at the first dip, so the axes would need to be offset by maybe 1 or 0.5 tick from bottom left to accomodate that. im not certain it can be made to look good, but maybe you'd like to give it a try? also svg would be preferable to png.

leeoniya:
@MUmarShahbaz added a logo here:

https://github.com/leeoniya/uPlot/blob/master/uPlot.svg

MUmarShahbaz:
Well damn 😅

PR #1060, opened 2025-05-20
Got rid of "wide" links in index.html

enderwiggum:
The anchor tags on index.html page were styled to be block elements (display: block), that caused the entire width of the page around a given anchor to be clickable... which has repeatedly annoyed me when trying to look at the demos / learn uplot.

Example: when focusing a browser window displaying the index.html by clicking in whitespace on the page, you accidentally trigger the link to whatever demo your pointer happens to be horizontally aligned with.

I removed the display:block from the anchor tags, wrapped said anchors in <div>, and gave each div the class 'demo'.  The page looks the same, but the links are now confined to the immediate space around the anchors.

### DIFF
--- a/demos/index.html
+++ b/demos/index.html
@@ -23,6 +23,10 @@
 			-webkit-tap-highlight-color: rgba(0, 0, 0, 0);
 		}
 
+		.demo {
+			padding: 0.5ex;
+		}
+
 		h1 {
 			font-size: 24px;
 		}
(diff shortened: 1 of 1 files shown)


leeoniya:
hey, thanks :)

instead of wrapping each link in a div, what do you think about wrapping all the links in a div like this:

```html
<div style="white-space: pre-line; line-height: 2em;>
  ...
</div>
```

Issue #922, opened 2024-02-05
Interactive canvas layer


Minardil:
Hello. I am planning to fork uplot to make it possible to move hovering to second canvas in order to improve performance when there are a lot of (200 and more series. Maybe we could discuss it and support this functionality in the library?

leeoniya:
i'm open to exploring this. i am sure updating one canvas element will be cheaper than updating 200 dom elements. i'd like to see some tests of how it impacts mem as well, not just cpu, in cases of 100 charts with 5 series each.

besides .u-over, there is also .u-under, as well as plans to add a clipped and unclipped region in https://github.com/leeoniya/uPlot/issues/876.

taking all this into consideration makes this not completely trivial. long term, i would prefer not to have two different strategies either.

Minardil:
Here is small demo. It takes time to render because there are a lot of series on each chart. https://minardil.github.io/uPlot/demos/lot-of-series.html. Changes in uplot are now dirty but I think it could be done as a plugin if there is an exposed method like u.draw(idx, opts: {canvasCtx, stroke, fill, width, etc})

leeoniya:
thanks, i'll take a look soon.

i did a quick test of the impact of creating multiple canvases and putting bunch of hover point dom elements inside.

performance profile of moving the mouse originally showed a lot of cost going to browser hit-testing, even though there are no hover css styles and no listeners attached to the points. i found a neat trick to reduce this by a lot:

https://stackoverflow.com/questions/41830529/optimizing-native-hit-testing-of-dom-elements-chrome/51534641#51534641

(but the main perf cost is not this tho. the main cost is the actual `.style.transform` mutation of each hover point)

```html
<!doctype html>
<html>
  <head>
      <title>many-can</title>
      <style>
        body {
          margin: 0;
        }

        canvas {
          background: pink;
        }

        .wrap {
            position: relative;
            display: inline-block;
            margin: 8px;
        }

        .overlay {
          width: 100%;
          height: 100%;
          position: absolute;
          opacity: 0;
          z-index: 100;
          left: 0;
          top: 0;
        }

        .pt {
            top: 0;
            left: 0;
            position: absolute;
            will-change: transform;
            pointer-events: none;
        }
      </style>
  </head>
  <body>
    <script>
      let count = 30;
      let series = 200;

      let width = 800;
      let height = 600;

      let pxRatio = devicePixelRatio;

      let ctxs = [];

      for (let i = 0; i < count; i++) {
[...]

Issue #431, opened 2021-01-11
use SVG for markers


leeoniya:
ideally, tiny, single-path SVGs where the path `d` attribute can be reused for both the svg elements of hover points as well as Path2D cached paths for canvas draw.

https://github.com/grafana/grafana/pull/30192

these can be in the core if they're small and visually distinctive:

- circle
- square
- triangle
- star
- plus
- asterisk

maybe:

- diamond (rotated square?)
- cross (rotated plus?)

e.g. tiny star:

```svg
<svg viewBox="0 0 260 245"><path d="M55 237L129 9l74 228L9 96h240"/></svg>
```

PR #1082, opened 2025-08-26
Point form customization

Eubehi:
Several things are done in this PR:

1. Give the users a handle to customize points drawn.
Today, the example to override the point form (draw-hooks.html) is not performant. In order to draw the custom point, the `points.show` function is overidden; this force the drawing of every point, regardless of the serie's size.
With this PR, the alignement of the points to be drawn is handled by the uPlot library, no need to override the `points.paths` function and copy the alignment logic.
2. Allow to change the Cursor form (based on a simple SVG path).
Maybe we could reuse the draw method, but SVG are nicely displayed and does not suffer of pixel blur.
3. Allow the forms to be visible in the legend (using the SVG defined for the cursor).

Some discrepancies remain:

- Cursor's size and width and point's size and width are defined in pixels, while the legend marker width is defined relatively to the SVG viewBox. It can be weird for the user.
- Since the marker is now handled by SVG, the dashed property is not handled the same as it was with plain CSS value.

This PR shows that pixel alignment is weird. For example, the Small CROSS of size 8 and width 1, when pixelAlign is set to 1, looks like more a diamond than a CROSS. I'll work on pixelAlign next.


Eubehi:
I'm marking this PR as ready to review. But in my mind, I wanted to open a discussion with you to cursor's form and legend markers. If you prefer, I'll mark it as a draft again.

leeoniya:
> I'm marking this PR as ready to review. But in my mind, I wanted to open a discussion with you to cursor's form and legend markers. If you prefer, I'll mark it as a draft again.

thanks for the PR, i will take a look soon.

however, i definitely prefer to discuss things like this in an issue/proposal first instead of getting a random code dump without prior context. it will help you better align your PRs with future changes and understand where things are headed that are not immediately obvious from the repo code or conversations.

https://github.com/leeoniya/uPlot/issues/431

i've been thinking long term about rendering markers on a dedicated canvas layer, which would fix the misalignment issues between DOM and canvas: https://github.com/leeoniya/uPlot/issues/922

i'm currently experimenting with fast svg pathstring marker rendering in https://github.com/leeoniya/uMarks

> I'll work on pixelAlign next.

if you attain an improvement, maybe put that into a separate PR (if it improves the existing hover point alignment)

leeoniya:
also, please leave the build files out of PRs for easier review. i will do a build commit when the PR is merged.

leeoniya:
> Today, the example to override the point form (draw-hooks.html) is not performant. In order to draw the custom point, the `points.show` function is overidden; this force the drawing of every point, regardless of the serie's size.

this may be true of the demo, but this callback does receive all the necessary arguments needed for conditional point visibility. it does require copying some code, though.

https://github.com/leeoniya/uPlot/blob/abadf17b94c1f6866ea7f3495988c5e4d5ad5ea3/dist/uPlot.d.ts#L830

leeoniya:
i've written a bit more about this here: https://github.com/grafana/grafana/issues/111040

Issue #924, opened 2024-02-12
Is there a way to reset scales programmatically?


trufflesprouts:
I'm trying to add a button to reset all scales similar to the way that double click works. Is there way to invoke the [built in implementation](https://github.com/leeoniya/uPlot/blob/5756e3e9b91270b303157e14bd0174311047d983/src/uPlot.js#L3252-L3264) of double click? or any clean way of doing this?

My current solution is saving the initial min and max in `_range` then setting the scale to be that value.

Another solution is emitting a double click event on `u-over` which seems a bit hacky.

leeoniya:
you can autorange/reset any scale by setting its min and max to `null`. the x scale is special, and resetting it will also reset all the y scales, assuming they have their original `range` fns. if you want to simulate what double-click does, you can do `u.setScale('x', {min: null, max: null})`

trufflesprouts:
i tried setting it to null but it doesn't work as expected https://jsfiddle.net/7pzqbdy1/61/
the second solution works

leeoniya:
let's leave this open. since i think null/null should work.

PR #1097, opened 2025-10-14
fix(types): Allow setScale to accept { min: null, max: null }

marcgallego:
Hi! 
First of all, let me thank you for this amazing library!

This PR introduces a minor change in the `setScale` type definition, allowing it to be called with `{min: null, max: null}`, as documented in #924.

The new type only allows null values if both arguments are null. Otherwise, I have seen that the axis may disappear in some edge case (such as `{min: 0, max: null}`).

If you find it reasonable, I can add some checks at the beggining of the `setScale` function to prevent this. The presence of special "numbers" such as `NaN` or `Infinity` also cause the same problem. However, I understand if you prefer to leave validation to the user to avoid additional runtime checks.


leeoniya:
thanks!

PR #1101, opened 2025-10-19
Multiple time series with different gaps (fixes #1100)

### LINKED ISSUE #1100: Multiple time series with different gaps
I would like to migrate from dygraph to uPlot, but this is currently blocking.

A real-life use case is when you want to display the prices of two financial instruments in a single plot. Their values change independently (at different times), and market availability is different as well.

## Example

I can easily render these plots with disabled `spanGaps`, as long as I show only one of the series in a single plot:

<table>
  <tr>
    <td>

| Time | A value |
| --- | --- |
| 1 am  | 10 |
| 3 am  | 20 |
| 5 am  | 30 |
| 6 am  | unavailable  |
| 8 am  | 50 |
| 9 am  | 40 |

</td>
    <td>
Image
    </td>
  </tr>
</table>

<table>
  <tr>
    <td>

| Time | B value |
| --- | --- |
| 2 am  | 30 |
| 4 am  | 40 |
| 5 am  | unavailable |
| 7 am  | 40 |
| 9 am  | 30 |

</td>
    <td>
Image
    </td>
  </tr>
</table>

When I merge them to be able to display in a single chart, I would like to have something like this:

<table>
  <tr>
    <td>

| Time | A value | B Value |
| --- | --- | --- |
| 1 am  | 10 | no change |
| 2 am  | no change | 30 |
| 3 am  | 20 | no change |
| 4 am  | no change | 40 |
| 5 am  | 30 | unavailable |
| 6 am  | unavailable | no change |
| 7 am  | no change | 40 |
| 8 am  | 50 | no change |
| 9 am  | 40 | 30 |

</td>
    <td>
Image
    </td>
  </tr>
</table>

However, I tried all combinations of `null` / `NaN` / `undefined`, but I am not able produce this result using uPlot.

Using dygraph, "no change" should be `null`, "unavailable" can be `NaN` or `undefined` and then will be rendered as expected:

Image

### Live demo

You can find and play with this example here: https://stackblitz.com/edit/vitejs-vite-tmyaqmsr?file=src%2Fmain.ts

## Explanation

When you want to display a time series with gaps, you can use nulls and keep `spanGaps` disabled, which works well. However, when you want to display two time series on the same chart, you have to merge them into a single dataset (table). Then you can use `null` to indicate that a given series does not update the value at a time, and use `spanGaps` to span over these gaps. But you need a way to represent the real gaps as well. Dygraph manages that by allowing `null` and `NaN`/`undefined` values in the dataseries. `null` means no value update for the given series, `NaN` or `undefined` means a missing value. In dygraph, `NaN` values always produce gaps, and `null` values are spanned or not based on the `connectSeparatedPoints` option (~`spanGaps` in the uPlot world).

kondi:
This is one proposed solution for issue #1100. Please check it to understand the context.

Possible approaches:

### 1. New `spanNulls` option

Keep current behaviour untouched, `spanGaps`: `true` or `false` are unchanged, `null`/`NaN`/`undefined` are handled the same way as before - as long as you do not use the freshly introduced `spanNulls` option. When you set `spanNulls` to `true` and keep `spanGaps` as `false` (the default value), uPlot will span nulls but will keep gaps at `NaN` or `undefined`. Like dygraph does when `connectSeparatedPoints` option is `true`. If you set `spanGaps` to true, uPlot will span all gaps, regardless of the `snapNulls` value (like before).

Risk: no real risk with backward compatibility, but could be confusing for somebody if they use both options together.

The pull request currently implements this first approach.

### 2. New values for `spanGaps`

Introduce new, non-boolean values for the `spanGaps` option. Could be numbers of an enum, and something like this:
 - `SpanGaps.DISABLED = 0` or just `false` (default value): do not span (like `false` previously)
 - `SpanGaps.ALL = 1` or just `true`: span over all gaps (like `true` previously)
 - `SpanGaps.ONLY_NULLS = 2`, new behaviour, similar to `connectSeparatedPoints: true` in dygraph

Risk: if somebody passed `spanGaps: 2` option before for whatever reason, it was considered trueish and handled `NaN`/`undefined` values differently than it would with this approach.

### 3. Never span over `NaN` or `undefined`

Change current `spanGaps` behaviour to never span over `NaN` or `undefined`, even when it is set to `true`. This is exactly what dygraph does when `connectSeparatedPoints` is `true`.

Risk: if somebody wants to span over `NaN` or `undefined`, it won't be possible anymore. 

## Live demo

You can find the live example of issue #1100 with my fix here: https://stackblitz.com/edit/vitejs-vite-8gtkovpo?file=src%2Fmain.ts

## Summary

I went with the safest option to ensure full backward compatibility, but I’m open to implementing the other approaches - or even a completely different solution.

If you want, I can add examples to the demos folder.


leeoniya:
hey @kondi, thanks for the PR!

what you're looking for is already possible using `alignGaps` option of the relevant pathbuilders (see convo in https://github.com/leeoniya/uPlot/issues/1100#issuecomment-3422204241).

maybe it makes sense to expose the `alignGaps` option on the Series interface (similar to `spanGaps`) so that it's not necessary to init the pathbuilders manually to control this :thinking:

leeoniya:
> maybe it makes sense to expose the `alignGaps` option on the Series interface (similar to `spanGaps`) so that it's not necessary to init the pathbuilders manually to control this :thinking: 

i tried this out locally and realized you'd still need to init the pathbuilders manually. i'm not sure i want to have two ways of doing the same thing. i'll probably still do it so the option is easier to discover (and it applies to all the same pathbuilders that support spanGaps). for uPlot v2 i'll have to think harder about where i want it, but the goal would be to make things more tree-shakeable, so many options might move to pathbuilder-specific opts anyhow.

leeoniya:
added in https://github.com/leeoniya/uPlot/commit/2ae22a59a0be2490aad6e9b8400306ad3b1e1960

PR #1117, opened 2025-12-09
update numAxisSplits to throw error instead of infinite loop

fastfrwrd:
We have had some instances recently where the min and max being equal crashes the browser in Grafana:

- https://github.com/grafana/grafana/pull/115030
- https://github.com/grafana/grafana/pull/109702

While we should not attempt to remediate this inside uPlot, it would be better to just throw if we can tell that's about to happen rather than running out of memory.

### DIFF
--- a/src/opts.js
+++ b/src/opts.js
@@ -599,6 +599,12 @@ export function numAxisSplits(self, axisIdx, scaleMin, scaleMax, foundIncr, foun
 
 	scaleMin = forceMin ? scaleMin : roundDec(incrRoundUp(scaleMin, foundIncr), numDec);
 
+	if (scaleMin === scaleMax)
+		throw new RangeError("scaleMin cannot be equal to scaleMax");
+
+	if (roundDec(scaleMin + foundIncr, numDec) === scaleMin)
+		throw new RangeError("foundIncr too small for scaleMin");
+
 	for (let val = scaleMin; val <= scaleMax; val = roundDec(val + foundIncr, numDec))
 		splits.push(Object.is(val, -0) ? 0 : val);		// coalesces -0
 


leeoniya:
hey paul, nice to see you :)

> While we should not attempt to remediate this inside uPlot

since uPlot has to handle flat data (e.g. single data points), there is already `min ~= max` remediation in the default scale ranging functions, e.g.: https://github.com/leeoniya/uPlot/blob/3a56bc6a03c0c07899d950254d9e5a4427d9835a/src/utils.js#L206-L271

this is why splits finding functions [and everything downstream from `scale.range`] expect that `min ~= max` situation is already handled early and in one place. is the situation that in Grafana we're customizing the `scale.range` option and not doing the same checks?

the example data in the added test seems to work okay with the built-in ranger: https://jsfiddle.net/z0xaq398/

```js
let opts = {
  width: 400,
  height: 170,
  scales: {
    x: {
      time: false,
    },
  },
  series: [
    {},
    {
      stroke: "red",
    },
  ],
};

let data = [
  [0, 1, 2, 3, 4],
  [2, 1.999999999999999, 2.000000000000001, 2, 2],
];

let u = new uPlot(opts, data, document.body);
```

while i agree that crashing and infinite loop is not a great outcome, it should be the responsibility of the customized options to uphold the same contract as the built-in functions (the ranging util functions are also statically exported for this purpose). i wonder if uPlot should additionally expose some finer-grained helpers to avoid too much copy-paste into any customized scale.range callbacks. and/or provide some kind of conformance tests that allow someone to validate that their own range functions handle all the same situations that uPlot's default ones already handle.

fastfrwrd:
> is the situation that in Grafana we're customizing the scale.range option and not doing the same checks?

Yeah, that's exactly right. The Sparkline is doing its own range configuration, mainly because it's user configurable in some situations like Table, and we keep running into cases where this happens. I wonder if there's a way to do the kind of check we do for auto-range to a configured range, and then maybe adjust what was provided if it's going to be a problem.

> i wonder if uPlot should additionally expose some finer-grained helpers to avoid too much copy-paste into any customized scale.range callbacks

I like that idea!

fastfrwrd:
one util that would be nice to have would be some way to get the implied `decimal` value, if we have to do this rounding out in Grafana ourselves, we could really use a way to know what precision uPlot is gonna round by if no decimal is sent in so we can get the same rounded values out. (I believe that's `guessDecimals` in grafana but would be better to call uPlot to make sure that we're always in sync)

leeoniya:
@fastfrwrd could you minimally extend the above jsfiddle to show how an exported decimals lookup would be used in a custom `scale.range`? i guess with the currently-problematic values?

PR #1121, opened 2026-01-10
perf(legend): reduce amount stress caused by too many legends

H4ad:
This PR will contain these main changes:
- using document fragment to avoid create elements directly in the dom
- using requestFrameAnimation + insert the legends & cursor points last
- remove ::after and ::before css to avoid costs of reflow

To test:

```diff
diff --git a/bench/uPlot-10M.html b/bench/uPlot-10M.html
index 5af53e2..ca855a6 100644
--- a/bench/uPlot-10M.html
+++ b/bench/uPlot-10M.html
@@ -109,7 +109,7 @@
 			let data1 = prepData(2e6);
 			let u1 = makeChart(data1, false);
 
-			let data2 = prepData(40_000, 50, 2);
+			let data2 = prepData(100, 10000, 2);
 			let u2 = makeChart(data2, true);
 
 			setTimeout(() => {
@@ -131,4 +131,4 @@
 			}, 2000);
 		</script>
 	</body>
-</html>
\ No newline at end of file
+</html>
```

Before:

image

After:

image

> Maybe in the future introduce a feature to virtualize or just render top 20 legends and add button to load more.

I found this while trying to optimize the https://github.com/SigNoz/signoz/issues/9784

> This is breaking change since it could break people that expected legends to be on DOM after new uPlot.

### DIFF
--- a/src/dom.js
+++ b/src/dom.js
@@ -49,14 +49,17 @@ export function setStylePx(el, name, value) {
 	el.style[name] = value + "px";
 }
 
-export function placeTag(tag, cls, targ, refEl) {
+export function placeTag(tag, cls, targ, refEl, frag) {
 	let el = doc.createElement(tag);
 
 	if (cls != null)
 		addClass(el, cls);
 
-	if (targ != null)
+	if (frag != null) {
+		frag.appendChild(el);
+	} else if (targ != null) {
 		targ.insertBefore(el, refEl);
+	}
 
 	return el;
 }
--- a/src/uPlot.css
+++ b/src/uPlot.css
@@ -1,7 +1,5 @@
 .uplot,
-.uplot *,
-.uplot *::before,
-.uplot *::after {
+.uplot * {
 	box-sizing: border-box;
 }
 
@@ -79,6 +77,7 @@
 .u-inline.u-live th::after {
 	content: ":";
 	vertical-align: middle;
+	box-sizing: border-box;
 }
 
 .u-inline:not(.u-live) .u-value {
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -589,6 +589,8 @@ export default function uPlot(opts, data, then) {
 			NULL_LEGEND_VALUES[k] = LEGEND_DISP;
 	}
 
+	let legendRowsFrag = null;
+
 	if (showLegend) {
 		legendTable = placeTag("table", LEGEND, root);
 		legendBody = placeTag("tbody", null, legendTable);
@@ -600,15 +602,20 @@ export default function uPlot(opts, data, then) {
 			legendHead = placeTag("thead", null, legendTable, legendBody);
 
 			let head = placeTag("tr", null, legendHead);
-			placeTag("th", null, head);
+			let headFrag = doc.createDocumentFragment();
+			placeTag("th", null, null, null, headFrag);
 
 			for (var key in legendCols)
-				placeTag("th", LEGEND_LABEL, head).textContent = key;
+				placeTag("th", LEGEND_LABEL, null, null, headFrag).textContent = key;
+
+			head.appendChild(headFrag);
 		}
 		else {
 			addClass(legendTable, LEGEND_INLINE);
 			legend.live && addClass(legendTable, LEGEND_LIVE);
 		}
+
+		legendRowsFrag = doc.createDocumentFragment();
 	}
 
 	const son  = {show: true};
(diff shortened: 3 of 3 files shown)


leeoniya:
> Maybe in the future introduce a feature to virtualize or just render top 20 legends and add button to load more.

that's actually an upcoming topic of discussion in Grafana. we have users with 2k+ series per panel, and the legend being in the dom becomes a performance issue, (even though Grafana uses its own React legend)

> using requestFrameAnimation + insert the legends & cursor points last
> This is breaking change since it could break people that expected legends to be on DOM after new uPlot.

that's my main concern with this approach, too.

H4ad:
> that's my main concern with this approach, too.

Can't we just emit an new event and make this a breaking change release?

leeoniya:
imo, rendering an insane amount of legend items is fundamentally useless. since human eyes are unlikely to accurately distinguish more than 20-30 colors, trying to slightly optimize the legend for hundreds+ series use case is not a great approach to fix both a performance and a UX issue.

virtualization is not really possible with a non-table legend which flows/wraps unpredictably, so that solution would only help the table legend case, which is not the default.

since uPlot does not currently have a tooltip (it may get one in v2.0), it kinda has to render all items to make it possible to see the raw values on hover. i want to make sure that any solution still allows a live legend to display the hovered value(s). a couple new options to paginate/show-all the legend in addition to being able to live-sort it by hovered value may work well enough for all use cases and be much faster than trying to optimize css/dom for a thousand series. :thinking:

leeoniya:
after giving this more thought, i'm going to do something more drastic.

we've also seen a request come up in Grafana for the tooltip to be limited but also show the series closest to cursor, so we need to be able to isolate and reorder those series as needed.

uPlot's legend is also its tooltip. and as a lib that specializes in rendering huge data sets, i think there needs to be more useful control over the tooltip & legend when applied to hundreds or thousands of series:

- filtering, like by proximity, and/or by removing null valued series at the cursor hover point
- sorting, like largest value first, or in the table legend case maybe by multiple properties besides the series value.
- limit/pagination, maybe virtualization.

supporting all of this will only be possible by moving the legend implementation to a vitual dom model; manual dom management is no longer realistic.

i'm going to start a branch here that begins the work of rewriting the legend using https://github.com/localvoid/ivi. once there is feature parity, i will add options for proximity-testing, sorting, filtering, and pagination.

i dont expect the lib size to increase by very much, since ivi is extremely optimized for both performance and bundle size. unlike react which is both slow and is multiple times the size of uPlot.

btw, thank you for the Luxon PR in grafana, we will get to it soon. the team that owns the code is currently working on some other things. replacing Slate and Monaco with codemirror being one of them :). https://github.com/grafana/grafana/pull/98505

H4ad:
Feel free to ping me for feedbacks or review while working on rewriting the legend, I'm really interested to see the improvements.

About filtering, maybe you can take a look at https://github.com/pingec/downsample-lttb, could help reduce amount of points in the chart and also filter out some legend that is not contributing to the chart directly.

leeoniya:
my thoughts on LTTB are here:

- https://news.ycombinator.com/item?id=46709415
- https://github.com/leeoniya/uPlot/pull/1025

however, LTTB is not applicable here because we're talking about filtering/sorting/limiting _series_, not _values_. this is to manage the UX of the legend/tooltip, not the plot itself.

leeoniya:
groundwork just merged: https://github.com/leeoniya/uPlot/pull/1144

H4ad:
Nice, from what I tested, the performance seems to be the same after migrating to ivi but now with the potential to unlock the new features, thanks for the hard-work!

Closing this one since it became stale.

leeoniya:
now that the internal legend contract is understood, it's easy to experiment while only touching 140 loc that live in the legend. 

im gonna evaluate swapping ivi for [roqajs](https://github.com/roqajs/roqa), it's looking to be about 2kb smaller and should be same perf or a bit faster.

https://krausest.github.io/js-framework-benchmark/current.html

leeoniya:
that PoC did not pan out, but https://github.com/leeoniya/uPlot/pull/1145 did.

let's see how this ages. but we'll always have a working/tested ivi implementation in our back pocket just in case.

H4ad:
I would pick the easier one to implement what the lib needs, that will produce less code possible, since the performance of each other are very close

We can save few kbs by choosing one but then introducing the same amount of more by writing more code to do the same functionality

PR #1123, opened 2026-01-22
update lightweight-charts bench demo for v5 of the library

SlicedSilver:
Hi, I came across this benchmark file for Lightweight Charts and noticed that it loads the latest version of the library but hasn't had the syntax updated for the v5 breaking changes. I've made minimal changes to get it working again, and also added the `minBarSpacing` setting so that all the datapoints are plotted within the visual viewport of the chart (as this appears to match what the other examples are doing).


leeoniya:
thanks!

PR #1125, opened 2026-02-19
Fix: When findGaps, do not use spread operator because it may cause Maximum call stack size exceeded

vbackeberg:
When setting large, interleaved datasets (~1M points), findGaps will crash due to the spread operator.

The proposed solution is to assign directly from findGaps.
It appears to be a simple fix and I could not think of any edge case that I overlooked.
I applied it to linear, stepped and spline, but only tested it for linear.

Since uPlot is capable of handling large number of gaps apart from this issue, I though this might be a valuable change.

### DIFF
--- a/src/paths/linear.js
+++ b/src/paths/linear.js
@@ -146,9 +146,9 @@ export function linear(opts) {
 
 			if (!series.spanGaps) { // skip in mode: 2?
 			//	console.time('gaps');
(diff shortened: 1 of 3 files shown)


leeoniya:
thanks! i'll do a follow-up to flag gaps like the linear one does.

PR #1137, opened 2026-07-01
Let user pick the stroke order

ad8e:
When my mouse hovers over a series, I'd like to bring it up above the other series. This patch lets the user choose the stroke order.

This is wandb: the hovered line is brought to the top and the other runs are translucent.

Screenshot 2026-07-01 at 2 35 04 PM

Here's my own chart: I increase the width of the line and make the other runs translucent, but there are enough runs to obscure the hovered line anyway.

Image

After this patch, the cerulean line is now visible:

Screenshot 2026-07-01 at 2 42 32 PM

The hovered line swaps at 60 fps as my mouse sweeps over the graph, so this method avoids reordering the data (which would be expensive).

If `self._zOrder` isn't the API you want (I know no javascript), other solutions would be fine too.

### DIFF
--- a/src/uPlot.js
+++ b/src/uPlot.js
@@ -1620,7 +1620,10 @@ export default function uPlot(opts, data, then) {
 				}
 			});
 
-			series.forEach((s, i) => {
+			// order the strokes by self._zOrder (a permutation of series indices) when set, so callers can lift series on top without reordering the series/data arrays.
+			// the order only affects stroking — all paths (incl. band fills) are built above, order-independently.
+			for (let i of self._zOrder || series.keys()) {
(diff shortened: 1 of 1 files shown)


leeoniya:
interesting, looks clean, thanks for keeping the PoC minimal.

are you expecting this to work together with the existing mouse and legend hover?

https://leeoniya.github.io/uPlot/demos/focus-cursor.html

i think making this the default behavior might make sense and would avoid the need to add it to the public API. if there is significant additional value in exposing control over this it might be best exposed as `series.zIndex` and you would simply loop over the series and call `u.setSeries(i, {zIndex: 5})` or something.

PR #1138, opened 2026-07-01
Draw _focus series on top

ad8e:
Suggested by https://github.com/leeoniya/uPlot/pull/1137#issuecomment-4860405109

> are you expecting this to work together with the existing mouse and legend hover?

I think so now. This patch reuses the `_focus` flag. I don't need arbitrary orders, but I do need to focus a few series simultaneously, because they are [smoothed curve, raw values, envelope] for a single run together.

Cost: an extra loop over the series, whether anything is focused or not.

I considered saving the _focus series to an array then drawing those after, to avoid two passes over the series. But the code is uglier that way.

The previous PR had no extra loop when nothing is focused, but responsiveness matters most when the focus is changing, so that's not important.

I'm not confident the build artifacts satisfy your customs; Claude handled the /dist folder.


leeoniya:
> but I do need to focus a few series simultaneously, because they are [smoothed curve, raw values, envelope] for a single run together.

how are you handling this, then? the existing focus behavior only handles a single focused series (both programatically via `setSeries()` and on cursor/legend hover):

https://github.com/leeoniya/uPlot/blob/0e5812c504430f5c804e0f993376d8999b26cc34/src/uPlot.js#L2560

are you manually setting `u.series[i]._focus` and then calling `u.redraw()` or something?

ad8e:
That's right: I manually set `_focus` and call `redraw(False)`. So my code, as currently based on this patch, only uses `_focus` to bring a series (with smoothed curve and envelope) to the top, and bypasses the existing machinery.

The reason I use custom highlighting logic is that the highlighted series links a few places: the same series in other charts, a list of runs, and some tooltips.

I also override the alpha properties of focus, and handle alpha myself (and line width). I tried telling Claude to piggyback on `focus`'s existing `alpha` handling, but it did not want to because focus's alpha handling is designed for only a single focused series.

Here's what it looks like:

Screenshot 2026-07-01 at 8 19 02 PM

ad8e:
Context: in ML, it's mandatory to render both raw values (to see NaNs/infs/spikes) and smoothed values, which can be compared between runs.

PR #1139, opened 2026-07-13
Add ink-uplot to Third-party Integrations

planadecu:
Adds [ink-uplot](https://github.com/planadecu/ink-uplot) under Third-party Integrations.

It's a terminal renderer for uPlot: it reuses standard browser uPlot config (`opts`/`data`, series, scales, dual axes, custom tick formatters) and draws the chart in the terminal via React Ink — as truecolor Unicode block art, or as real inline images through the kitty, sixel, and iTerm2 graphics protocols. Useful for CLI dashboards, time series, and live/streaming plots.

A terminal integration alongside the existing framework and Python wrappers.


leeoniya:
what sourcery is this!? 😅

planadecu:
Terminal UIs don't have a standard charting lib. In my project we widely use uPlot in the web, with this we don't need to port the charts we can still reuse the uPlot web definitions for the termina app.

PR #1142, opened 2026-09-13
fix infinite loop in numAxisSplits for big-magnitude flat ranges

### LINKED ISSUE #827: Infinite loop in numAxisSplits
I develop a company-internal tool that displays plots using uPlot. For a specific data set, I observe uPlot blocks when displaying the page, because it is stuck in an infinite loop in `numAxisSplits` in opts.js.

```
    for (let val = scaleMin; val <= scaleMax; val = roundDec(val + foundIncr, numDec))
        splits.push(Object.is(val, -0) ? 0 : val);    // coalesces -0
```

On Firefox, the console displays "Uncaught out of memory". On Chrome, the error page is displayed: Aw, Snap! Error code: STATUS_BREAKPOINT.

I tried to create a minimal example to reproduce the infinite loop, but I failed to do that. This is because the initialization of the uPlot object is very dynamic in my tool, and I could not find a hardcoded version that has the same behavior. I can say something about the dataset itself though: it happens only when the plot line is flat and with a high value. I can observe the problem with the dataset [100000000000000,100000000000000], but not with [100000000000000,100000000000001].

When I step into the code with a debugger, I can see that `val`, `scaleMin` and `scaleMax` all equal to 100000000000000. `foundIncr` is 1e-8 and `numDec` is 8. The problem is that `val + foundIncr` evaluates to 100000000000000, and thus `val` never exceeds `scaleMax`.

I found a workaround, but I don't know if it is correct. I observed that in the caller code (`axesCalc` in uPlot.js), the variable `_space` becomes `Infinity` in this situation. I saw that we exit early if `__space` is 0, so I added a check for Infinity in addition to 0.

```
    if (_space == 0)
        return;
```

```
    if (_space == 0 || _space == Infinity)
        return;
```

cpruijsen:
`numAxisSplits` in `src/opts.js` can hang the page. Its loop advances with
`val = roundDec(val + foundIncr, numDec)`, and when `numDec` is too small for the increment, the
rounded result comes back equal to the previous `val`. The condition `val <= scaleMax` stays true
forever and the tab locks up rather than rendering a wrong axis.

Two changes, in the order the failure happens.

`numIntDigits` in `src/utils.js` computed the digit count with the branchless int32 absolute value
`(x ^ (x >> 31)) - (x >> 31)`. Both shifts coerce to a signed 32 bit integer, so any magnitude at or
above 2^31 wraps and the digit count comes back wrong, which is what produces the `numDec` that
cannot represent the increment. It now takes `abs(x)` and only truncates with `| 0` while the value
is inside int32 range, using the float directly above it.

The loop then carries a `prevVal` guard and stops when a step fails to advance. That is a backstop
rather than the fix: with the digit count corrected the loop terminates on its own, but a stalled
step should end the loop rather than the session, since the remaining causes of a non advancing step
are floating point ones that are hard to enumerate.

Fixes #827


leeoniya:
thanks, i have fixed this differently. and more thoroughly i hope.

https://jsfiddle.net/769uwrq1/

